PR #14, opened 2025-10-29
Cautious Weight Decay support

sippycoder:
Cautious Weight Decay (https://arxiv.org/pdf/2510.12402) is shown to have small performance gains consistently over current weight decay update. They apply weight decay only to parameter coordinates whose signs align with the optimizer update.

This PR adds Cautious Weight Decay support for `Muon`, `AdamW`, and `Lion` optimizers. 

Enable cautious weight decay with `cautious_wd` directly in Muon initialization 
```
optimizer = Muon(
    param_groups,
    distributed_mesh=device_mesh,
    lr=learning_rate, 
    mu=0.95,
    weight_decay=weight_decay,
    cautious_wd=True,                      # applies cautious weight decay
    epsilon=1e-8,
    nesterov=True,
    adjust_lr="rms_norm",
    use_triton=True,
)
```

or apply `cautious_wd` in param_group:
```
param_groups.append(dict(params=muon_params, cautious_wd=True))
```

### DIFF
--- a/dion/scalar_opts.py
+++ b/dion/scalar_opts.py
@@ -168,9 +168,6 @@ def adamw_update_foreach(
         # Reference: https://arxiv.org/pdf/2510.12402
         coeff = lr * weight_decay
         
-        # decay_masks = [
-        #     (m_div * x >= 0).to(dtype=x.dtype) for x, m_div in zip(X, M_div)
-        # ]
         decay_masks = torch._foreach_mul(X, M_div)
         decay_masks = torch._foreach_sign(decay_masks)  # {-1, 0, 1}
         decay_masks = torch._foreach_add(decay_masks, 1)  # {0, 1, 2}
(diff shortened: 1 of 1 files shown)


sippycoder:
You're right! `_foreach_minimum` is a catch! Nice one. I updated the logic. I'll test it and compare on my 64B model run.

dion/muon.py
```diff
@@ -560,14 +570,24 @@ def muon_update_post_orthogonalize(
...
     Inputs and outputs should be lists of regular Tensor, not DTensor.
     This is a separate function for compatibility with torch.compile().
     """
-    # Apply weight decay
-    torch._foreach_mul_(X, 1 - base_lr * weight_decay)
+    if cautious_wd:
+        # Apply cautious weight decay: only where update and parameter signs align
+        # Reference: https://arxiv.org/pdf/2510.12402
+        coeff = base_lr * weight_decay
+        decay_masks = [(u * x >= 0).to(dtype=x.dtype) for x, u in zip(X, U)]
```
byronxu99: Change this to the foreach version

sippycoder: Done.

sippycoder: I already tested it, I just forgot to add to this repo.

dion/scalar_opts.py
```diff
@@ -44,8 +45,19 @@ def adamw_update(
     # Adjust learning rate to include bias correction 1
     adj_lr = lr / bias_correction1
 
-    # Apply weight decay
-    X.mul_(1 - lr * weight_decay)
+    if cautious_wd:
+        # Compute update direction (pre-LR) for CWD mask
+        update_dir = M / denom
```
anhnguyen-namitech: Just a passerby, is this needed? Denom is > 0 anyway so it does not matter.

byronxu99: I think it's fine either way. Putting `M / denom` here may not be necessary but it makes the meaning more clear. For efficiency we will pretty much always use the foreach version, which does not encounter this issue.

PR #25, opened 2026-01-16
Dion2 with shard-independent update route added

kwangjunahn:
This PR adds a shard_independent argument to the Dion2 optimizer, giving users control over how top-K row selection interacts with FSDP sharding.

##  1. Sharding-independent row selection for FSDP
Dion2 selects a subset of rows (top-K by L1 norm) for orthogonalization at each step. When parameters are sharded across devices, there's a design choice: should selection happen on local shards or on the full matrix?

### Options

- `shard_independent=False` (default)
  - Behavior: Each shard independently selects its top-K rows based on local momentum
  - Pros: Only communicates selected submatrices → reduced communication cost
  - Cons: Selection depends on sharding configuration.
  - **Note: Empirically, this dependency doesn't appear to affect training outcomes**

- `shard_independent=True`
  - Behavior: All-to-all gathers the full matrix, then selects top-K rows "globally" on each matrix
  - Pros: Algorithm behavior is identical regardless of sharding configuration
  - Cons: Requires communicating full matrices (communication cost of this path matches that of Muon)
  
## 2. Row-wise Selection Only

We now restrict Dion2's selection to row-wise selection (i.e., `select_dim=-2`). This ensures the selection semantics match "neuron-wise" selection, since linear layers have shape (output, input) — each row corresponds to one output neuron.
Previously, the code would choose between row or column selection based on matrix shape or sharding configuration. This change makes the behavior more interpretable and consistent.

### DIFF
--- a/configs/debug.yaml
+++ b/configs/debug.yaml
@@ -1,0 +1,42 @@
+
+# — Model —
+model_dim: 768
+n_layer: 12
+n_head: 6
+sequence_length: 128 
+
+# — Batching & Training — 
+num_iterations: 3000
+batch_size: 1024
+device_batch_size: 256
+
+# — Learning‐rate schedule —
+warmup_ratio: 0.0
+warmdown_ratio: 0.2
+
+# — Validation & Checkpointing —
+val_loss_every: 125
+val_tokens: 10485760 
+
+# — Weights & Biases logging —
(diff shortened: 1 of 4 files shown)


JohnLangford:
This is super out of date, so closing for now.

PR #31, opened 2026-03-07
Add polar express

### LINKED ISSUE #26: Add Polar Express Muon
According to the following paper, using Polar Express instead of Newton-Schultz in Muon leads to improvements in the validation loss.
https://arxiv.org/abs/2505.16932

It was also used in Karpathy's NanoChat repo to train a GPT-2 equivalent model for less than $100 (https://github.com/karpathy/nanochat/discussions/481)

It might be worthwhile to add it here.

**Official implementation:** https://github.com/NoahAmsel/PolarExpress/blob/main/polar_express.py

segyges:
Should resolve #26, polar express implementation matching the Keller Jordan and Karpathy implementation including a triton implementation speedup included in the existing triton newton-schulz implementaiton

### DIFF
--- a/dion/muon.py
+++ b/dion/muon.py
@@ -9,6 +9,7 @@ from torch.optim.optimizer import Optimizer, ParamsT
 from typing import Callable, Generator, List, Optional, Tuple, Union
 
 from .newton_schulz_triton import newton_schulz_triton
+from .polar_express import polar_express, polar_express_triton
 from .opt_utils import (
     AsyncRuntime,
     AsyncTask,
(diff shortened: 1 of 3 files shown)


segyges:
@johnlangfor

> @segyges I have yet to find some setting where this is advantageous either computationally or statistically. Do you have ideas here?

I haven't A/B tested them, so I actually have no idea here. It's plausible to me that the benefit actually ended up being contingent on or tightly coupled to the oddities of Keller's modded-nanogpt, which would not be the MOST surprising result but would be a little bit surprising, because the mathematical motivation doesn't seem like it's obviously coupled to their or any other setup.

PR #36, opened 2026-03-29
Add ARO (Adaptively Rotated Optimization) optimizer

JohnLangford:
## Summary
Implements the ARO algorithm ([arxiv:2602.09006](https://arxiv.org/abs/2602.09006)), which replaces Newton-Schulz orthogonalization with an adaptive rotation policy based on QR decomposition.

**Algorithm per step:**
1. Update momentum: M = β·M + (1-β)·G
2. Rotate into previous frame: R_prev^T @ M
3. Apply base optimizer f_t (row_norm or sign)
4. Cross-alignment + QR → new rotation R
5. Rotate with new R, apply f_t → update direction U
6. Apply: W -= η·U

**Implementation:**
- Subclasses `DistributedOrthoBase` for distributed setup, step orchestration, Lion/AdamW tasks
- R stored in float32 for QR stability, adding O(m²) memory per param
- DDP megabatch distributes QR+matmul across ranks via all-gather
- Base optimizer functions: `row_norm` (default), `sign`
- FSDP not supported (R requires full row dimension)

**Tested:**
- Single-GPU: rotation stays orthogonal (3.6e-7 error after 3 steps)
- DDP 2×H100: exact cross-rank agreement (0.00 diff)

## Test plan
- [x] Single-GPU smoke test (mixed shapes, megabatch, both base_opt)
- [x] DDP cross-rank consistency (2×H100)
- [ ] Training loss comparison vs Muon/AdamW on a real model

### DIFF
--- a/dion/__init__.py
+++ b/dion/__init__.py
@@ -1,3 +1,4 @@
+from .aro import ARO
 from .dion import Dion
 from .dion import DionMixedPrecisionConfig
 from .dion_simple import Dion as DionSimple
--- a/dion/aro.py
+++ b/dion/aro.py
@@ -296,8 +296,10 @@ def _shifted_cholesky_qr(A: Tensor) -> Tensor:
     # Shift proportional to the Frobenius norm of A
     shift = A.norm() ** 2 * 1e-7
     G.diagonal(dim1=-2, dim2=-1).add_(shift)
+    # Release cached memory right before cusolver call — the matmul above
+    # re-fills the cache, so the earlier empty_cache() is not sufficient.
+    torch.cuda.empty_cache()
     # Upper Cholesky: G = R^T R, then Q = A @ R^{-1}
-    # Same pattern as dion.py's orthogonalize()
     R, info = torch.linalg.cholesky_ex(G, upper=True)
     if (info != 0).any():
         # Fallback: Householder QR for ill-conditioned inputs
--- a/dion/aro.py
+++ b/dion/aro.py
@@ -1,6 +1,5 @@
 import math
 import torch
-import torch.distributed as dist
 from collections import defaultdict
 from torch import Tensor
 from torch.distributed import ProcessGroup
@@ -8,7 +7,7 @@ from torch.distributed.tensor import DTensor
 from torch.optim.optimizer import ParamsT
 from typing import Callable, Generator, List, Optional, Tuple, Union
 
-from .megabatch_base import DistributedOrthoBase
+from .megabatch_base import DistributedOrthoBase, megabatch_orthogonalize_async
 from .opt_utils import AsyncTask, to_local
 from .muon import adjust_lr_spectral_norm, adjust_lr_rms_norm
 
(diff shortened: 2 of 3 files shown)


JohnLangford:
Shifted Cholesky QR also fails — same `cusolverDnCreate(handle)` error, now at `torch.linalg.cholesky_ex`. The CUDA 13.0 bug affects cusolver handle creation itself, not a specific decomposition. Any `torch.linalg.*` operation that touches cusolver will fail.

The workaround needs to avoid cusolver entirely. Options:
1. **CPU fallback** for the Cholesky path — `cross.cpu()` → cholesky → triangular solve → `.to(device)`. The cross matrix is `[per_rank, m, m]`, QR/cholesky is not the bottleneck
2. **Pure matmul implementation** — Gram-Schmidt or iterative refinement using only `torch.mm` (which uses cublas, not cusolver)
3. **Newton-Schulz iteration** to approximate orthogonalization of the cross matrix (same as what Muon/dion2 already use — no linalg calls)

Option 3 is arguably the cleanest since dion already has Newton-Schulz infrastructure and it stays fully on GPU.

JohnLangford:
**Re-tested 2026-04-29 on the current cjob image** (`nvcr.io/nvidia/vllm@sha256:19a3cff5…`, torch 2.10 nv25.12, CUDA 13.1) on an 8×B200 / 1B mixformer / climbmix smoke. Recording the findings before re-closing so the next person doesn't repeat the loop.

### Throughput is the blocker, not correctness

ARO converges and FSDP2 works. But the QR path is the bottleneck:

| smoke run | QR backend | step time | opt% of step |
|---|---|---:|---:|
| `phinext-jl-aro-test-smoke` (1B/100M) | CPU `torch.linalg.qr` | 6.37s | 93.5% |
| `phinext-jl-aro-test-smoke-gpu-qr` (1B/100M) | GPU cusolver | crash on step 0 | — |
| `phinext-jl-aro-test-smoke-magma` (1B/100M) | MAGMA preferred + CPU fallback | 7.31s | 94.5% |

For comparison on the same setup, Muon is 0.45s/step (5.5% opt%) and MuonSphere is 0.57s (25.6%). ARO is **14–16× slower** on the CPU-QR path.

### Why GPU QR fails on this image

```
RuntimeError: cusolver error: Unknown cusolver error number, when calling
`cusolverDnCreate(handle)`. ...
You are running using the stub version of cusolver.
```

The image ships a **cusolver stub** — `cusolverDnCreate` returns a "Unknown cusolver error", so anything routed through cusolver fails immediately. This is a different failure mode from the original B200/CUDA-13.0 corruption that motivated the CPU-QR fallback in the first place; on this image cusolver isn't usable at all.

`torch.backends.cuda.preferred_linalg_library("magma")` is silently a no-op because MAGMA isn't installed in the image either. So `torch.linalg.qr` has no working GPU backend.

### What would unblock this

Three options, roughly in order of effort:

1. **Add MAGMA to the cjob image.** A Dockerfile package add. The MAGMA-preference line I tried (commit `9c6b30d`) becomes effective the moment MAGMA is present.
2. **Find / build a cusolver-real variant of the cjob image.** NVIDIA's vllm image apparently ships only the stub at this tag.
3. **Re-implement ARO's QR cusolver-free.** Cholesky-of-Gram + triangular-solve, all on GPU via cuBLAS only. This is what `_shifted_cholesky_qr` (line 288) tries to do but it currently routes through `torch.linalg.cholesky_ex` which also goes through cusolver.

Closing the PR again until one of those lands. The branch (`dev/aro`) is preserved with two attempted-fix commits on top:

- `3fabc72` — try GPU QR (failed: cusolver stub)
- `9c6b30d` — prefer MAGMA, CPU fallback (worked but still on CPU path)

Reverting those out of the branch would just throw away the empirical record; leaving them on the closed branch so the next attempt can pick up where this left off.

PR #62, opened 2026-04-16
Add per-group num_heads option for per-head Newton-Schulz

JohnLangford:
## Motivation

Attention Q/K/V/gate projections are typically stored as a single fused 2D `nn.Linear` weight of shape `(num_heads * head_dim, in_features)`, but semantically each head is an independent `(head_dim, in_features)` matrix. Running Newton-Schulz on the fused matrix orthogonalizes a single large matrix that blends information across heads.

To get per-head orthogonalization today, models have to store the weight as a 3D tensor with a custom multilinear Linear layer — an invasive model-side change (extra forward-pass reshape, per-rank checkpoint-format change, TP plan gap).

This PR lets users keep `nn.Linear` 2D and get per-head NS by setting `num_heads` on the parameter group instead.

## Change

Adds a per-param-group `num_heads` option (default: `None`, no change in behavior). When set, the optimizer internally views each 2D weight as a batched 3D tensor `(num_heads, head_dim, in_features)` for orthogonalization:

```python
param_groups = [
    dict(params=attn_proj_params, num_heads=config.num_attention_heads),
    dict(params=other_matrix_params),
    dict(params=vector_params, algorithm="adamw"),
]
optimizer = Dion2(param_groups, lr=lr, distributed_mesh=mesh)
```

- Newton-Schulz runs per-head (leverages the existing `flatten=False` batched NS path).
- `spectral_norm` / `rms_norm` LR adjustment uses the per-head matrix shape `(head_dim, in_features)`.
- With FSDP sharding dim 0 along head boundaries (requires `num_heads % world_size == 0`), each rank's shard holds whole heads — NS runs locally without the all-to-all that assembling the fused matrix would need.

## Scope

- Implemented for Muon, Dion2, and NorMuon. All three share `_prepare_head_split` in `DistributedOrthoBase`. NorMuon's per-neuron variance buffer is reshaped alongside the parameter.
- Axis-0 heads only. O-projection stores heads on axis 1 and would need a different treatment; not in this PR.

## Tests

New `TestNumHeads` class in `test_optimizers.py`:

- **Parity**: `Muon` / `Dion2` / `NorMuon` with `num_heads=N` on a 2D weight produce updates identical to the same optimizer run on the equivalent 3D-stored weight (tested across several steps with matching gradients). Covers standard Muon, Muon+Nesterov, Dion2 at fractions 0.5 and 1.0, and NorMuon.
- **Validation**: `num_heads` that doesn't divide dim 0 raises `ValueError`; `num_heads > 1` on a non-2D param raises.
- **Megabatch**: multiple same-shape params with `num_heads` set are batched together correctly.

All 45 tests in `test_optimizers.py` pass (44 pre-existing + 8 new under `TestNumHeads`).

## Notes for reviewers

- The NS step for 3D tensors was unblocked by #59 (Dion2) and already supported by Muon/NorMuon's `flatten=False` path — this PR is essentially a thin wrapper that exposes it without requiring a model-side layout change.
- In-place updates on the reshaped view propagate through shared storage, so momentum and variance buffers stay 2D while being processed as 3D each step.
- `normuon_normalization_stacked`'s per-matrix Frobenius norm (`dim=(-2, -1)`) and per-neuron variance (`dim=-1`) broadcast correctly over the extra leading head dim — the kernel itself needs no changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,7 +243,7 @@ param_groups = [
 
 ### Per-Head Newton-Schulz for Attention Projections
 
-Attention Q / K / V / O / gate projections are typically stored as a single 2D `nn.Linear` weight of shape `(num_heads * head_dim, in_features)`, but semantically each head is an independent `(head_dim, in_features)` matrix. Orthogonalizing the fused matrix blends information across heads — often not what you want.
+Attention Q / K / V / gate projections are typically stored as a single 2D `nn.Linear` weight of shape `(num_heads * head_dim, in_features)`, but semantically each head is an independent `(head_dim, in_features)` matrix. Orthogonalizing the fused matrix blends information across heads — often not what you want.
 
 You can ask Dion2, Muon, or NorMuon to run Newton-Schulz independently per head without changing the model layout by setting `num_heads` on the parameter group:
 
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ You can find the following optimizers:
 1. [Optimizers](#optimizers)
 1. [Building Parameter Groups](#building-parameter-groups)
    * [Example Code](#example-code)
+   * [Per-Head Newton-Schulz for Attention Projections](#per-head-newton-schulz-for-attention-projections)
 1. [Distributed Training Configuration](#distributed-training-configuration)
    * [Flattened Meshes](#flattened-meshes)
    * [Device Mesh for Muon](#device-mesh-for-muon)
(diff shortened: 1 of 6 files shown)


alint77:
I have tested this on nanoPLM and the improvement was in the ~1e-3 range, the param count was tiny (70M, 8H, 96Hdim) tho so it might work better on bigger models.

JohnLangford:
Thanks.  Testing here seems to be going well also.  Merging.

PR #63, opened 2026-04-17
Fused scalar_opts adamw: replace foreach with torch._fused_adamw_ (10x fewer kernel launches) 

alint77:
## Problem

Dion's `adamw_update_foreach` spent ~138 ms of CPU time per step in our pretraining loop, while the actual GPU work for it only takes ~13 ms. That's a big gap, and it showed up clearly in profiler traces: `NorMuon.step` was CPU-bound.

The cause is that `torch._foreach_*` isn't really fused on the CPU side — each call dispatches one `aten::<op>` per tensor. With ~150 AdamW params and ~6 foreach ops per step, that's ~900 aten dispatches every step. `@torch.compile` helps group the GPU kernels but can't collapse the per-tensor Python/C++ dispatch. For comparison, swapping in `torch.optim.AdamW(fused=True)` dropped the same work to ~43 ms of CPU.

## Fix

I wrote a Triton multi-tensor-apply (MTA) kernel that does the same math as `torch._fused_adamw_` (see [`fused_adam_utils.cuh`](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/cuda/fused_adam_utils.cuh)) but processes all params in a single kernel launch. Cautious weight decay is a `tl.constexpr` branch inside the same kernel so there's one code path.

Pointer arrays and partition metadata are cached across steps, keyed on each tensor's `data_ptr()`. I originally used `id()`, but `param.data` and `DTensor.to_local()` return a fresh Python wrapper each access — so an id-keyed cache misses every step and grows unboundedly under FSDP. `data_ptr()` stays stable across wrapper re-creation. Pinned-memory staging keeps even cache misses from syncing the stream. Gradient pointers get re-validated each step since `set_to_none=True` allocates fresh grad buffers.

## Results

Per `Optimizer.step`, averaged over 4 profiled windows:
TORCH baseline: AdamW params handled by a separate torch.optim.AdamW(fused=True)

| | DION (pre-fix) | TORCH baseline | DION (post-fix) |
|---|---|---|---|
| `NorMuon.step` CPU | **138 ms** | 40 + 3 ms | **40 ms** |
| `cudaLaunchKernel` | 320 | 29 | **22** |
| `aten::lerp` | 298 | 0 | **0** |
| GPU kernel time | 12.6 ms | 11.7 ms | 10.6 ms |
| AdamW GPU kernels | ~600 | ~30 | **1** |

Microbenchmark on realistic Dion AdamW shapes (150 × 1-D bf16 tensors): ~95 ms/step → ~0.5 ms/step.

Correctness: 20 tests in [`dion/tests/test_adamw_mta.py`](dion/tests/test_adamw_mta.py), all passing. Bit-exact against `torch.optim.AdamW(fused=True)` at bf16, within 1 ulp at fp32. Covers CWD on/off, fp32/bf16/fp16, weird shapes, cache hit/miss, FSDP-style fresh wrappers, and end-to-end `NorMuon` + `Muon` integration.

**BEFORE:**
Screenshot 2026-04-17 at 23 43 18

**AFTER:**
Screenshot 2026-04-17 at 23 43 29


JohnLangford:
Nice result and the profiler write-up is convincing — a real CPU-side dispatch bottleneck, solved without changing the caller contract. A few substantive concerns and one structural alternative:

## Biggest alternative worth considering: just call `torch._fused_adamw_`

PyTorch already ships a multi-tensor-apply AdamW kernel (`torch._fused_adamw_` / `torch.optim.AdamW(fused=True)`). Your own numbers put its CPU cost at 43 ms vs your kernel's 40 ms — essentially a tie. The main reason this PR exists is that CWD isn't in the fused op. A simpler shape would be:

- Non-CWD path: call `torch._fused_adamw_` directly, skip the whole cache + metadata + Triton kernel.
- CWD path: keep a small foreach impl, or apply CWD as a post-step correction.

That removes ~250 lines of kernel + metadata caching, drops the Triton dependency, and inherits whatever PyTorch does in the future (mixed-dtype state, CUDA graph capture, compile passes). Worth considering before merging a parallel kernel the project will now own.

## Correctness concerns

1. **`data_ptr()` cache key is not sufficient.** PyTorch's caching allocator reuses addresses. If a tensor is freed and another allocated at the same address with a different `numel`, you get a cache hit with stale `numels` / `block_to_tensor` / `block_to_chunk_start` — silent OOB writes. The test in `test_cache_hit_across_fresh_wrappers` only validates the happy case where storage is genuinely stable. At minimum, include `numel` in the key: `(data_ptr, numel)` per tensor.

2. **No contiguity check.** The kernel does `x_ptr + offs` with linear offsets. A non-contiguous view (slice, transpose, `.data` on a sharded FSDP1 flat param, …) would silently corrupt memory. The old `_foreach_*` path was stride-aware. Add `assert t.is_contiguous()` on cache miss, or fall back.

3. **Cache never evicts.** Under workloads where param addresses rotate — FSDP2 reshard, checkpoint reload, EMA swap, `optimizer.load_state_dict` — `_mta_cache` grows unboundedly. Each miss adds ~MB-scale CUDA int tensors. Consider a weakref-based cache keyed on the `torch.Tensor` storage, or an LRU cap.

4. **Silent dtype-uniformity regression.** The old foreach path tolerated `M.dtype != V.dtype` (explicit casts on `G` into each). New kernel requires all of X/M/V to share dtype. Dion's current state init uses `zeros_like(param)` so this happens to hold, but it's now a silent invariant — assert it at the Python boundary with a clear error rather than letting a `KeyError` on the dtype lookup fire deep inside.

5. **Triton dependency is now mandatory** for this code path (old path had a pure-torch body). Raising on import-less environments is a small regression for CPU testing / non-Triton GPU stacks. Keeping the old `_foreach_*` impl as a fallback is ~30 lines.

## Testing

- `test_bit_exact_step1` uses `atol=0, rtol=0` for bf16 vs `AdamW(fused=True)`. That's a bold invariant to hold across Triton / CUDA / GPU arch revisions — likely to become flaky. A small tolerance (e.g. `atol=1`, rtol 0 on the int bits, or just `atol=1e-3`) would be safer.
- No test for the two silent-corruption modes above: (a) data_ptr reuse across steps, (b) non-contiguous input. Worth adding.
- No distributed / DTensor test despite the PR description's emphasis on FSDP motivating the `data_ptr()` choice.

## Minor

- `adamw_update_foreach` is now a misnomer — it's neither foreach nor per-tensor dispatch. Rename (or alias) for readability.
- `_pinned_int_tensor_to_device` rebuilds `ptrs_g` on every step when `set_to_none=True` (so the ~1–2% savings from the pinned path shows up every step, not only on cache miss). Fine, just worth noting in the docstring that "cache hit" still does one H2D per step.
- `BLOCK_SIZE=1024` is hard-coded; a `triton.autotune` or at least a comment on why 1024 would help future readers.

Overall: the performance delta is real and the kernel math looks right, but I'd strongly consider whether `torch._fused_adamw_` gets you 95% of this with a fraction of the surface area, and tighten the cache before this ships.

alint77:
I noticed this adamw unfused issue a while back for our nanoPLM project, I just routed adamw param group to torch.optim.adamw, but I wanted to have CWD on adamw too so that's why I made this PR now. also since I know this repo is being used for some very large scale runs by big labs, I wanted to contribute the fix since the savings are going to be massive with such a small change. 

regarding "apply CWD as a post-step correction" I think that is a fair point, I just wanted to have a unified implementation for now by pretty much copying the internals of torch's adam (with triton instead of cuda), since already there is triton dep for NS, I thought it'd be ok. 

edit: I will try the cwd post-step correction path now to see if it works. will update soon

alint77:
@JohnLangford please check the update, it now applies cwd as post-step to avoid triton and adam reimplementation.

JohnLangford:
Much cleaner — thanks for taking the `_fused_adamw_` route. +596/-66 → +175/-65, same perf characteristics, zero custom kernel to maintain. This is the right shape for the problem.

Quick second pass on the update:

## Verified good

- **Shared `step_t` across all params is safe.** I checked empirically: `_fused_adamw_` does not mutate `state_steps` (at least for fp32 scalar step), so passing `[step_t] * n` has no write-race, and the bias correction uses `step_t` directly — matching Dion's original `1 - beta1**step` semantics. No off-by-one.
- **CWD correction math is right**, including the sign==0 edge case (old path applied WD there; new path's `(signs < 0)` skips the undo, so the fused op's unconditional decay stands — same behavior).
- **Non-mutation of `state_steps` means the `.fill_(step)` each call is fully idempotent** regardless of PyTorch version.

## Minor remaining

1. **CWD path still pays foreach overhead.** The post-correction is `clone` + 3 foreach ops + 1 element-wise cast. That's cheap if CWD is on scalar groups only (biases/LN), but if someone enables `cautious_wd` on a large-param AdamW group you're partially back in the dispatch-cost regime you were trying to fix. Worth a one-line doc note: CWD reintroduces ~4 aten dispatches per param-list, so the perf win is biggest when CWD is off or on small groups.

2. **Non-CWD bf16 test uses `atol=1e-2, rtol=1e-2`** vs `torch.optim.AdamW(fused=True)` — that's loose for what should be numerically identical (same underlying kernel). If there's drift, it's likely from the fp32 vs int64 step-tensor dtype difference; worth narrowing the tol to catch a real regression, or documenting the source of the drift.

3. **`torch._fused_adamw_` is a private (underscored) API.** It's very stable de facto, but an explicit fallback to the old foreach body on `AttributeError` would be cheap insurance — and restore CPU-path coverage, which the old code had.

4. **`_step_tensor_cache` is module-global and never cleared.** Fine in practice (at most one entry per device), just noting it for symmetry with the earlier cache review — bounded, unlike the previous MTA cache.

Overall LGTM once the CWD perf-note / tolerance tightening are addressed. Nice turnaround.

JohnLangford:
This is real John (I've been experimenting with a PR review bot).   Thanks for this :-)  It makes no measurable difference on my workload, but seems like an obviously good idea while being more concise.

PR #71, opened 2026-05-01
Add: MuonH (Hyperball optimizer) + add option for 'neuron' or 'short_axis' normalization for norMuon and MuonH

alint77:
Both the optimizer and "short_axis" normalization inspired by https://github.com/KellerJordan/modded-nanogpt/pull/273. Big thanks to @WhenWen. 

[MuonH writeup](https://psychedelic-sunstone-851.notion.site/Fantastic-Pretraining-Optimizers-and-Where-to-Find-Them-2-1-Hyperball-Optimization-2e924306e6f280e7a5ffee00eb40a0dd)

only 3-4% slower on opt.step gpu time compared to normuon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,7 @@
 This repository provides efficient implementations of orthonormal optimizers for distributed ML training.
 You can find the following optimizers:
 * [Muon](https://kellerjordan.github.io/posts/muon/)
+* [MuonH](https://tinyurl.com/muonh)
 * [Dion2](https://arxiv.org/abs/2512.16928) and [Dion](https://arxiv.org/pdf/2504.05295) (Dion is a legacy optimizer; we recommend using Dion2)
 * [NorMuon](https://arxiv.org/abs/2510.05491) 
 
@@ -15,6 +16,7 @@ You can find the following optimizers:
 1. [Quick Start](#-quick-start)
 1. [Introduction](#introduction)
 1. [Optimizers](#optimizers)
+   * [MuonH](#muonh)
 1. [Building Parameter Groups](#building-parameter-groups)
    * [Example Code](#example-code)
    * [Per-Head Newton-Schulz for Attention Projections](#per-head-newton-schulz-for-attention-projections)
@@ -53,7 +55,7 @@ pip install git+https://github.com/microsoft/dion.git
 Then in your code, you can use:
 
 ```python
-from dion import Dion2, Muon, NorMuon, Dion
+from dion import Dion2, Muon, MuonH, NorMuon, Dion
 ```
 
 Please carefully go through this readme for detailed instructions on using our optimizers. There are major differences compared to PyTorch built-in optimizers, such as `Adam`/`AdamW`.
@@ -146,12 +148,12 @@ The practical effectiveness of orthonormal optimizers was first demonstrated by 
 
 Our current implementations support the following parallelization techniques:
 
-| Parallelization    | Dion | Dion2 | Muon | NorMuon |
-|--------------------|------|-------|------|---------| 
-| Single device      | Yes  |  Yes  | Yes  |   Yes   |
-| PyTorch DDP        | Yes  |  Yes  | Yes  |   Yes   |
-| PyTorch FSDP2      | Yes  |  Yes  | Yes  |   Yes   |
-| PyTorch FSDP2 + TP | Yes  |  No   | No   |   No    |
+| Parallelization    | Dion | Dion2 | Muon | MuonH | NorMuon |
+|--------------------|------|-------|------|-------|---------| 
+| Single device      | Yes  |  Yes  | Yes  |  Yes  |   Yes   |
+| PyTorch DDP        | Yes  |  Yes  | Yes  |  Yes  |   Yes   |
+| PyTorch FSDP2      | Yes  |  Yes  | Yes  |  Yes  |   Yes   |
+| PyTorch FSDP2 + TP | Yes  |  No   | No   |   No  |   No    |
 
 For faster performance, these optimizers will process parameters in batches and interleave multiple batches to overlap compute with communication.
 
@@ -159,8 +161,33 @@ We include optimizer implementations in the `dion/` directory of this repo.
  
 * `dion.py`: High-performance version of Dion. Depending on how each batch of matrices is sharded, we select the best communication patterns to compute Dion's orthonormal update. All-reduce operations may be split into reduce-scatter and all-gather across the batch dimension to more efficiently distribute work and avoid redundant computation.
 * `muon.py`: High-performance version of Muon. For sharded matrices, all-to-all communication is used to simultaneously unshard and distribute a batch of matrices. For replicated matrices, Muon will distribute work across all devices and all-gather final results.
+* `muonh.py`: Muon with a hyperball update. After Newton-Schulz orthonormalization (and optional update normalization), each matrix update is scaled relative to that matrix's Frobenius norm and projected back to its initial Frobenius norm.
 * **`dion2.py`**: High-performance implementation of Dion2, using a similar all-to-all communication pattern for distributed orthonormalization. Only an α-fraction of the momentum matrix is communicated and orthonormalized, significantly reducing both communication overhead and computation cost.
 * `normuon.py`: A variant of the Muon optimizer that introduces neuron-wise normalization to improve stability and convergence efficiency, modified to take similar arguments as `muon.py`. See [the paper](https://arxiv.org/abs/2510.05491) for more details.
+
+### MuonH
+
+MuonH combines Muon's momentum + Newton-Schulz direction with a hyperball projection step. Conceptually, the optimizer keeps each matrix parameter on a Frobenius sphere: after forming the Muon-style update direction, MuonH moves along that direction and projects back to the matrix's initial Frobenius norm. This implementation follows [Fantastic Pretraining Optimizers and Where to Find Them 2.1: Hyperball Optimization](https://tinyurl.com/muonh).
+
+Compared with Muon, MuonH has a few important usage constraints and semantics:
+
+* **Non-zero initialization is required.** MuonH initializes its hyperball radius from the parameter norm on the first optimizer step, so a zero-initialized matrix will raise an error.
+* **Hyperball state is per matrix.** For ordinary 2D weights this is one radius per parameter. For packed matrix tensors such as MoE expert weights of shape `(n_experts, d_out, d_in)`, MuonH treats each trailing 2D matrix independently when `flatten=False`, so each expert gets its own momentum / normalization / hyperball radius.
+* **`flatten=False` is the right mode for packed experts.** A tensor of shape `(..., d_out, d_in)` is treated as a batch of matrices. This is the intended mode for MoE expert MLP weights and other native 3D matrix batches.
+* **`flatten=True` changes the semantics.** With `flatten=True`, MuonH flattens the parameter to one 2D matrix before orthonormalization and hyperball projection. Use this for convolution kernels only when you explicitly want the whole tensor treated as one matrix.
+* **`num_heads` is not supported.** MuonH deliberately rejects `num_heads > 1`. If you want per-head MuonH behavior, represent heads as explicit matrices, e.g. a native 3D tensor of shape `(num_heads, head_dim, in_features)` or separate `Parameter`s.
+* **Optional normalization matches NorMuon-style update normalization.** `normalization=None` disables it, `normalization="neuron"` normalizes along the last dimension, and `normalization="short_axis"` normalizes along the shorter matrix axis.
+* **Matrix params do not use decoupled weight decay.** As with Muon, embeddings, LM heads, biases, and normalization parameters should remain on AdamW or Lion parameter groups.
+
+Example:
+
+```python
+optimizer = MuonH(
+    param_groups,
+    lr=0.01,
+    normalization="short_axis",
+)
+```
 
 We also provide some reference implementations:
 
@@ -172,17 +199,19 @@ We also provide some reference implementations:
 
 ## Building Parameter Groups
 
-Unlike typical PyTorch optimizers (e.g. `Adam`/`AdamW`), Dion and Muon require separating your model's parameters into different groups (same in spirit as [Modula](https://docs.modula.systems/)). These orthonormal optimization algorithms are only applicable to two-dimensional matrix weights. Non-matrix parameters require a different scalar optimizer algorithm (element-wise updates) and may also use a different learning rate. We currently support Lion and AdamW.
+Unlike typical PyTorch optimizers (e.g. `Adam`/`AdamW`), Dion and the Muon-family optimizers require separating your model's parameters into different groups (same in spirit as [Modula](https://docs.modula.systems/)). These orthonormal optimization algorithms are intended for matrix weights. For Muon, MuonH, and NorMuon, a parameter with shape `(..., d_out, d_in)` may also be treated as a batch of matrices when that matches the model semantics. Non-matrix parameters require a different scalar optimizer algorithm (element-wise updates) and may also use a different learning rate. We currently support Lion and AdamW.
 
 The details of parameter grouping are dependent on model architecture and implementation. Therefore, we leave it up to you to categorize your model's parameters and create the necessary parameter groups.
 
-* In transformer models and many other neural networks, most parameters are `nn.Linear` layers with two-dimensional weight matrices. These parameters should use Dion or Muon. A shape-dependent learning rate scale factor will be automatically applied for each matrix.
+* In transformer models and many other neural networks, most parameters are `nn.Linear` layers with two-dimensional weight matrices. These parameters should use Dion, Muon, MuonH, or NorMuon. A shape-dependent learning rate scale factor will be automatically applied for each matrix.
 * Biases in `nn.Linear` layers (if used) are one-dimensional vectors, which must be placed into a separate parameter group from the weight matrices. Use Lion or AdamW.
 * Normalization layers (e.g. `nn.LayerNorm`, `nn.RMSNorm`) may contain vectors of learnable weights. Use Lion or AdamW.
 * Embedding layers (e.g. `nn.Embedding`) are stored as 2D tensors, but should be treated as a collection of 1D vectors using Lion or AdamW. (Warning: using Dion here will run without error, but will give poor performance.)
 * Unembedding layers (e.g. LM head) are typically implemented as a `nn.Linear` layer, but shoud also be treated as a collection of 1D vectors. Furthermore, they should use a **smaller scaled learning rate**. It is very important to manually identify this layer and place it into its own parameter group, as it is otherwise indistinguishable from weight matrices!
 (Warning: using Dion here will run without error, but will give poor performance.)
-* Convolution layers typically use parameter tensors with 3+ dimensions. These are currently not supported for Dion. Support for convolution layers in Muon is experimental, and can be enabled with the option `flatten=True` to automatically flatten them to 2D matrices when computing the optimizer update.
+* Packed expert weights such as MoE MLP parameters with shape `(n_experts, d_out, d_in)` are a natural fit for Muon, MuonH, and NorMuon with `flatten=False`, since each expert is treated as an independent matrix.
+* Convolution layers typically use parameter tensors with 3+ dimensions. These are currently not supported for Dion. Support for convolution layers in Muon-family optimizers is experimental and generally requires `flatten=True` so the whole kernel tensor is treated as one 2D matrix.
+* MuonH requires non-zero matrix initialization. Do not assign zero-initialized output projections or other zero-initialized matrices to MuonH.
 
 We summarize the above in this table. Let `d_in` be the input dimension of the unembedding layer. In transformer language models, this is the base dimension of the model.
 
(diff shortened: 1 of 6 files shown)


JohnLangford:
This is John's reviewbot.

Solid contribution, hyperball math looks right, and the test for Frobenius-norm preservation directly verifies the core claim. Comments below in priority order.

### Correctness / robustness
- **`hyperball_radius_initialized` survives state-dict round-trips only because it's a Python `bool`**. Checkpointing libraries that round-trip optimizer state through anything stricter than `pickle` (e.g. some safetensors-style flatteners) will drop it and the next step will silently re-initialize the radius from the *current* (already-trained) parameter norm — i.e. you'll lose the original radius. Storing this flag as a `torch.tensor(0/1, dtype=torch.bool)` (or just inferring "uninitialized" from `state["hyperball_radius"] == 0`) makes resume robust.
- **Zero-init parameters hard-crash on the first step** (`test_zero_radius_rejected`). Output projections initialized to zero (a fairly common Muon practice) will hit this. Consider: warn + fall back to the first nonzero post-Newton-Schulz update norm, or document the requirement loudly in the docstring.
- **First-step `dist.all_reduce` for radius init is synchronous and outside the `AsyncTask` machinery** (`_initialize_hyperball_radii`, line ~176). It blocks the caller before the first task is yielded — fine because it's once, but easy to fold into the existing async generator if you ever care about the first-step latency.
- **`_get_shard_info` is not specialized for MuonH the way it is for NorMuon** (which rejects last-dim sharding). The new normalization paths handle reductions over sharded axes correctly, so this is probably intentional, but worth a comment or test confirming a last-dim-sharded MuonH+`neuron` actually trains.

### Performance
- **Two sequential all-reduces in `muonh_update_post_orthogonalize_async`** (one for `update_sq_sums`, one for `candidate_sq_sums`). These can be fused into one round by sending `[|U|², ⟨X,U⟩]` and computing `|candidate|²` in closed form: `|candidate|² = r² − 2·lr·r·⟨X,U⟩/|U| + lr²·r²`. Halves the comm count of the post-step on FSDP2.
- **`@torch.compile(fullgraph=True, dynamic=False)` will recompile per shape group** (different `N` in the stacked batch) and per `red_dim` value (`-1` vs `-2`). With many distinct module shapes this can balloon compile cache and warmup. NorMuon's existing helper omits `dynamic=False` for this reason — consider the same.
- **`normuon_short_axis_normalization_async` is *not* `@torch.compile`-d** while the neuron sibling is. Likely a meaningful per-step gap on the wide-matrix path. Splitting around the (optional) async all-reduce the way `muonh_normalization_async` does it would let you compile the pure-tensor halves.

### Design / DRY
- **MuonH duplicates ~80% of NorMuon**: state init, normalization, head-split, shape-group iteration. Hyperball is conceptually `NorMuon update direction` + `Frobenius-sphere projection`. A more elegant factoring would expose the projection as a post-step plug-in (or add a `hyperball=True` flag to NorMuon) and delete `muonh.py` entirely. As written, every future fix to NorMuon's mega-batch path has to be mirrored into MuonH.
- **`_canonical_normalization` is duplicated** in `muonh.py` and `normuon.py` with subtly different acceptance rules (MuonH accepts `None`/`"None"`, NorMuon does not). Move to `megabatch_base.py` and standardize.
- **`_fro_norms_async` is dead code** — defined, never called.
- **`hyperball_eps=1e-10` default**: that's well below `bfloat16` underflow; the param dtype-cast in `_muonh_step_compiled`/`_muonh_project_compiled` may collapse it to 0 in bf16. Bumping to `~1e-7` (or staying in fp32 through the scale calc) would be safer.

### Backward compatibility
- The NorMuon variance-buffer dtype shifts from `torch.zeros_like(param[..., 0:1])` to `torch.zeros(..., dtype=local.dtype)`. Equivalent today, but worth a one-liner that old checkpoints with `variance_neuron` re-validate by shape and silently reset on mode change. That silent reset (e.g. switching `neuron` → `short_axis` mid-run) is probably the right call but should be documented.

### Tests
- All tests are single-GPU. Given how much of this PR is FSDP2 sharding/all-reduce logic, at least one DTensor-sharded smoke test (matrix-sharded on dim -2 with `short_axis`, ideally also dim -1 for MuonH) would catch the kind of bug the inner-product fusion above would introduce.
- `test_short_axis_matches_neuron_for_tall` compares the bf16-precision `(U*U).mean` neuron path against the fp32 `square().sum()/n` short-axis path — `rtol/atol=1e-5` is tight; may flake on non-Hopper hardware.
- A `num_heads + short_axis` test would be cheap and exercises the post-head-split variance shape branch.

alint77:
Thanks for the review. Here is what I changed.

## Correctness / robustness

**`hyperball_radius_initialized` checkpoint robustness.** Now stored as a CPU `torch.bool` tensor. It survives tensor-only state-dict round-trips. CPU placement avoids a device sync on the steady-state check.

**Zero-init params.** Kept as a hard error. The radius has to come from somewhere, and a "first nonzero update norm" fallback would silently change behavior. Tightened the MuonH docstring to call this out: matrix params must be non-zero at construction, and zero-init output projections will hit it.

**First-step `dist.all_reduce` synchronous.** Left as-is. It runs once and folding it into the async machinery did not seem worth the complexity.

**`_get_shard_info` policy.** MuonH's reductions are correct under last-dim sharding, so the looser policy is intentional. I added a multi-process DTensor smoke test (see Tests below) that exercises `shard_dim=-1` with both `neuron` and `short_axis` to confirm.

## Performance

**Two sequential all-reduces in the post-step.** Fused into one round. The post-step now sends `[|U|², |X|², ⟨X,U⟩]` in a single all-reduce and computes `|candidate|² = |X|² − 2·s·⟨X,U⟩ + s²·|U|²` analytically. The hyperball step + projection collapse into one fused multiply-subtract. Halves the post-step comm count on FSDP2.

**`@torch.compile(... dynamic=False)`.** Dropped `dynamic=False` from all five compiled helpers in MuonH. Matches NorMuon's existing pattern and avoids per-shape recompiles.

**`normuon_short_axis_normalization_async` not compiled.** Now compiled. Split into `_normuon_short_axis_sum_sq_compiled` and `_normuon_short_axis_finish_compiled` around the async all-reduce, mirroring how MuonH does it.

## Design / DRY

**`_canonical_normalization` duplicated.** Moved to `megabatch_base.canonical_normalization` with an `allow_none: bool` flag. MuonH passes `allow_none=True`, NorMuon passes `False`. One implementation, one error message format.

**`_fro_norms_async` is dead code.** Deleted.

**`hyperball_eps=1e-10` and bf16.** Left at `1e-10`. Tracing the actual code: `update_norms.clamp_min(eps)` runs on a fp32 tensor (sqrt of a fp32 sum), and the resulting scale is cast to the param dtype only at the end. eps never enters bf16, so there is no underflow risk.

**MuonH duplicates ~80% of NorMuon.** Agreed in principle. What I wanted to do here was to have a working, fast implementation of MuonH, If you prefer it to be folded into Muon or Normuon as an option, it can be done. 

## Backward compatibility

**Variance buffer silent reset.** Added a docstring on `NorMuon._get_or_initialize_variance` explaining that the buffer is keyed by shape, so resuming a checkpoint and changing the `normalization` mode silently re-initializes the EMA. This is intentional.

## Tests

**Single-GPU coverage gap.** Added `TestMuonHDTensor`, a multi-process DTensor smoke test using `mp.spawn` with 2 NCCL ranks. It parametrizes over five sharding configs: batch-shard, matrix-shard at `dim -2` and `dim -1`, each with `neuron` and `short_axis`. Each config runs three optimizer steps and verifies (a) Frobenius radius preservation and (b) cross-rank parameter consistency.

**`test_short_axis_matches_neuron_for_tall` rtol/atol.** Loosened from `1e-5` to `rtol=1e-3, atol=1e-4`. The two paths are not bitwise identical (`neuron` reduces in param dtype, `short_axis` reduces in fp32 then casts back), and the new tolerances absorb that plus 3-step accumulation without flaking on non-Hopper hardware.

**`num_heads + short_axis` test.** Added four parity tests in `TestNumHeads`: `MuonH` and `NorMuon` × default and `short_axis`. Each compares the 2D-with-`num_heads` path against the explicit 3D path.

JohnLangford:
This is John's reviewbot. Substantive feedback on the latest revision (db2216b).

**Praise**
- The fused post-step is genuinely elegant: packing `[|U|², |X|², ⟨X,U⟩]` into a single all-reduce and using the closed form `|X − sU|² = |X|² − 2s⟨X,U⟩ + s²|U|²` saves a comm round and avoids materializing the candidate. Worth keeping.
- DTensor smoke test covering 5 (shard_dim, normalization) combinations is the right level of coverage for FSDP2 paths that no in-process test can reach.
- Lazy variance-buffer allocation keyed by `tuple(shape)` cleanly handles the `'neuron'` → `'short_axis'` shape divergence on wide matrices.

**Substantive concerns**

1. **`num_heads` + multi-rank radius semantics are untested and ambiguous.** When `num_heads is not None`, `_create_ortho_tasks` sets `process_group=None` and `shard_dim=None`, and `_initialize_hyperball_radii` / `muonh_update_post_orthogonalize_async` both skip their all-reduce. That means with FSDP2 + `num_heads`, each rank computes a per-rank-local Frobenius radius (collapsed over its own heads via `flatten(1).sum(-1)`) and projects to it. Depending on intent, this is either (a) "each rank preserves its own shard's initial norm" or (b) a bug — the docstring just says "projected back to its initial Frobenius norm" without disambiguating. The existing parity test is single-process, and the DTensor test doesn't combine `num_heads` with `Shard(0)`. Please add a test for that combination and pin down the intended semantic in the docstring.

2. **Per-original-param ball under `num_heads`.** Even within a single rank, head-split params are stacked as `[N, num_heads_local, head_dim, K]` and the radius/sums use `flatten(1).sum(-1)` → one scalar per *original* param, scaling the whole 3D tensor uniformly. The ortho step is per-head but the hyperball is per-param. Probably intentional but worth a one-line note next to the `flatten(1)`, since the orthogonalization treats heads independently and a reader might expect the same here.

3. **MuonH inherits dim‑(−1) sharding without override.** `NorMuon._get_shard_info` rejects `sharded_tensor_dim == ndim − 1`; MuonH does not. The DTensor test exercises `(-1, 'neuron')` and `(-1, 'short_axis')`, which is good — but the `'neuron'` reduction axis equals the sharded axis there, and the all-reduce in `muonh_normalization_async` is the only thing that makes that correct. Add an inline comment near the `shard_dim_neg == red_dim` check explaining that this is what differentiates MuonH from NorMuon's restriction; otherwise a future refactor may copy NorMuon's guard back over.

4. **Radius is captured at first `step()`, not at `__init__`.** The docstring's note ("Matrix parameters MUST be non-zero at optimizer construction. The hyperball radius is initialized from the parameter's Frobenius norm on the first step") is internally inconsistent: construction-time vs. first-step-time matter when an LR scheduler / warmup / param re-init runs between them. Either snap the radius eagerly in `__init__` (would need a sharded-aware norm there too), or update the docstring to say "first step" consistently — current wording reads as if it's both.

5. **Synchronous `bool((radii <= eps).any().item())` on every first-step path.** Forces a host-device sync. Only once per param group, so cost is negligible, but it's the only sync in the otherwise-async pipeline. Either accept and comment, or fold the check into the compiled path with a sentinel/NaN fallback.

6. **Variance-buffer silent re-init on mode change.** The NorMuon docstring documents that `'neuron' → 'short_axis'` for a wide matrix silently re-initializes the EMA on resume. That's defensible, but a `warnings.warn(...)` the first time the shape mismatch fires would save someone a debugging session — the silent path is too quiet for an optimizer state change.

7. **Backward compat: existing checkpoints without `normalization` key.** `defaults` now includes `normalization`, and `_canonical_normalization(group["normalization"])` is called in `_create_ortho_tasks`. PyTorch's `load_state_dict` *does* re-merge defaults into groups, so this should work, but it's worth a one-line resume test (save → load → step) for NorMuon to guard against future regressions.

**Smaller points**
- `_canonical_normalization(group["normalization"])` is called every step in `_create_ortho_tasks` even though the value was already canonicalized into defaults at `__init__`. Skip it — reuse the stored string.
- The double-wrapping (`canonical_normalization` public, `_canonical_normalization` private wrapper in each module) is dead weight after the dedup. Just call `canonical_normalization(x, allow_none=True/False)` at the use site.
- Test port: `29500 + abs(hash((shard_dim, str(normalization)))) % 500` is non-deterministic across runs (PYTHONHASHSEED). Use a deterministic hash (e.g., `hashlib.md5(...).digest()[0]`) so collisions across parallel CI jobs are reproducible.
- README lists Muon / Dion2 / NorMuon but not MuonH. Add a one-liner.
- `weight_decay` is in `MuonH`'s `defaults` but the docstring says MuonH matrix params don't use decoupled WD. Either drop the kwarg from MuonH's signature (param groups using AdamW/Lion can set it via group dict) or annotate the docstring more loudly so users don't expect it to apply to matrix params.
- `_local_square_sums` has a dead `stacked.ndim > 1` branch — `torch.stack(list_of_scalar_tensors)` always yields ≥1D, so the `else` is unreachable for the call sites.

**Elegant alternative — closed-form is already in.** The natural simplification I would have suggested (one fused all-reduce of `[|U|², |X|², ⟨X,U⟩]` + closed-form |X−sU|) is exactly what the latest commit landed. Nothing further to add on the post-step.

LGTM after (1) is clarified — the others are improvements but non-blocking.

JohnLangford:
In  a first test, I see very modest improvement over normuon in the 1-3 millinats range.  It is however essentially free.

alint77:
I will fix the num_heads issue tomorrow so it's ready to merge. 

I haven't had access to H100 nodes to run bigger tests, but on my limited testing on a toy 20M 512,1536,8L,4H encoder model, the results were really promising: 

Screenshot 2026-05-02 at 02 40 26

JohnLangford:
Across 1B/1B and 1B/10B runs I'm seeing millinat scale gains.   It's very small, but fairly consistent and cheap.

JohnLangford:
(The very minor effect that we are observing here may be due to running with things fairly tuned.)

alint77:
Did you by any chance also do a lr sweep with adjust_lr=None ?

JohnLangford:
I tried a bit higher and a bit lower than my current sweet-spot LR.  Both were worse.  I'm using spectral norm and generally find the perf there a bit better than what is possible with 'None'.

alint77:
Follow-up changes on top of the current PR:

- Removed `MuonH` support for `num_heads > 1`.
  The previous behavior mixed per-head Newton-Schulz with a shared hyperball projection, which is not a coherent `MuonH` semantic. `MuonH` now raises for `num_heads > 1`, while `num_heads=1` remains a no-op like the other optimizers. I just don't think it's worth the added complexity at this stage.  

- Fixed `MuonH` semantics for native 3D+ matrix batches such as MoE expert weights.
  `MuonH` now tracks hyperball radii per matrix when `flatten=False`, instead of using one radius for the whole packed tensor. For a parameter shaped like `(n_experts, d_out, d_in)`, each expert now gets its own radius / projection, matching the unit that Newton-Schulz and optional normalization already operate on.

- Kept `flatten=True` behavior as the explicit "treat the whole tensor as one matrix" mode.
  This preserves the intended convolution-style path while making the default `flatten=False` path coherent for packed expert tensors.

- Cleaned up the distributed post-step/radius code accordingly.
  The dead `num_heads`-specific `MuonH` plumbing was removed, and the remaining radius reduction path is now only the matrix-sharded one.

- Updated tests.
  Added rejection coverage for `MuonH` with `num_heads > 1`, preserved the `num_heads=1` no-op behavior, and added parity coverage showing that packed 3D expert tensors match separate 2D expert parameters under `MuonH`.

- Updated the README.
  Added a dedicated `MuonH` section, documented the non-zero init requirement, clarified `flatten=False` vs `flatten=True`, documented the packed-expert/MoE behavior, and explicitly noted that `num_heads` is not supported for `MuonH`. The README now also references the Hyperball / MuonH writeup as the basis for this implementation.

JohnLangford:
This is John's reviewbot. Follow-up on commit 82a6841 (good progress).

**Resolved from prior review**
- `num_heads` semantic ambiguity now raises explicitly. Cleaner than disambiguating; the per-rank-radius issue I flagged for `num_heads + multi-rank` is moot under the new contract. Tests cover the rejection path and `num_heads=1` no-op.
- Per-3D-batch radius (`shape[:-2]` per slice) is consistent with the per-head orthogonalization, and `test_3d_experts_match_separate_params` is a direct contract test that packed 3D ≡ N separate 2D params. Exactly the right test.
- README updated with MuonH.

**One new substantive issue: `flatten=True` on 3D+ params is shape-broken**

`_muonh_local_sums_compiled` always reduces with `sum(dim=(-2,-1))` regardless of `flatten`, but `_local_square_sums` and `_get_or_initialize_hyperball_radius` correctly branch on `flatten`. For a 3D param `(B, M, K)` with `flatten=True`:

- `_local_square_sums` collapses to `[N]` (one radius per param) ✓
- `_muonh_local_sums_compiled` returns `[3, N, B]` (per-batch sums)
- `radii` is `(N,)` (scalar per param), `sums[0]` is `(N, B)`
- `s_uf = adjusted_lr * radii / u_norm` ⇒ `(N,) / (N, B)` — broadcasts as `(1, N) / (N, B)`, which silently produces wrong values when `N == B` and crashes otherwise.

`extra = X_stacked.ndim - 1 if flatten else 2` is also wrong-shape downstream — at flatten=True the post-step needs to reduce all batch+matrix axes into the per-param scalar, not preserve a batch axis. No test exercises this path (the new 3D test uses `flatten=False`), so it'd silently land. Suggested fix: thread `flatten` into `_muonh_local_sums_compiled` and `sum` over `tuple(range(1, ndim))` when set; or just disallow `flatten=True` for ndim ≥ 3 with a clear error (matching how `flatten=True + num_heads` is rejected). The simpler latter option is consistent with how MuonH's hyperball semantic is now defined per-2D-slice.

**Still open from prior review (non-blocking)**
- `_canonical_normalization(group["normalization"])` still called every step in `_create_ortho_tasks` even though defaults are canonicalized at `__init__`. Skip the per-step call.
- Synchronous `bool((radii <= eps).any().item())` in radius init — only fires once, fine to leave.
- Variance silent re-init on `'neuron' ↔ 'short_axis'` for wide matrices is documented but no `warnings.warn`. Cheap to add and would prevent a debugging session on resume.
- Test port still uses `abs(hash((shard_dim, str(normalization)))) % 500` — non-deterministic across PYTHONHASHSEED. Use `hashlib.md5(...).digest()` or fixed offsets.

**Independent validation context**
Ran a pre-merge sanity at 1b/10b on climbmix (microsoft/aifsdk pipeline, NorMuon path with `normalization='neuron'`) — train CE 2.396 vs 6-run control mean 2.400 (Δ −0.004), val_bpb on climbmix/web/code all within 0.001 of control. So the NorMuon-side `'neuron'` default really is shape-compatible with existing checkpoints, as the docstring claims. Did *not* sweep `normalization='short_axis'`; the test coverage there is single-process only. If you have a multi-rank validation-loss number for `short_axis` I'd be interested.

LGTM after the `flatten=True + 3D` fix.

alint77:
Pushed follow-up fixes for the flatten path, on both MuonH and NorMuon

- Fixed MuonH(flatten=True) so the hyperball post-step reduces over the fully flattened matrix, and added a 3D-vs-explicit-2D parity test.
- Fixed the same semantic mismatch in NorMuon(flatten=True) by making the variance/normalization path flatten-aware, with matching parity coverage.
- Also cleaned up a related MuonH sharded short-axis normalization shape issue that surfaced while adding the tests.
Targeted tests now pass for both MuonH and NorMuon.

JohnLangford:
This is John's reviewbot. Quick follow-up on 7468d3b — the `flatten=True + 3D` fix.

**Resolved**
- `_muonh_local_sums_compiled` and `_local_square_sums` now branch on `flatten` (`tuple(range(1, ndim))` when set), so the per-N scalar `radii` and per-N sums are shape-consistent for ndim ≥ 3 + flatten=True. The crash/wrong-broadcast I flagged is gone.
- Both `_get_or_initialize_variance` implementations now compute `red_dim` against the *flat* shape, not the original. That's a semantic fix on top of the broadcast fix — the old `param.shape[-2] >= param.shape[-1]` would have picked the wrong axis for short_axis on a 3D-flattened-to-2D matrix.
- The `shard_dim_pos == 0 → -2 else -1` mapping in the async normalization paths correctly translates a global-axis shard into the post-flatten axis.
- `test_flatten_3d_matches_explicit_2d` (both MuonH and NorMuon, all normalizations) is the right contract test for the flatten semantic — packed-3D-with-flatten ≡ pre-flattened-2D.

**Still worth tightening (non-blocking)**

1. **Sharded + flatten=True is uncovered.** Single-process flatten tests + sharded non-flatten tests exist independently, but no test exercises both. The `shard_dim_pos → flat_axis` mapping (line ~430 in muonh.py and the matching block in normuon.py) is the kind of two-step coordinate logic that's easy to get wrong on the next refactor. A 2-rank smoke covering `(shard_dim=0, flatten=True)` and `(shard_dim=1, flatten=True)` on a 3D param would lock the behavior down.

2. **Variance shape derived from local shape, not synchronized.** Each rank computes `flat_shape = (local.shape[0], prod(local.shape[1:]))` independently. With even FSDP2 sharding this is fine (every rank gets the same flat dims), but with uneven shards (rare but possible) different ranks could pick different `red_dim` values for `short_axis`, silently. Worth either an assert or a comment that even sharding is required.

3. **Variance silent re-init on mode change** is still un-warned (NorMuon docstring documents it; MuonH inherits the behavior). A one-line `warnings.warn` the first time the shape mismatch fires would prevent a real debugging session on resume — cheap to add.

4. **Test port still uses `abs(hash(str)) % 500`.** PYTHONHASHSEED randomization breaks reproducibility; replace with `hashlib.md5(...).digest()[0]` or a fixed table.

5. **Per-step `_canonical_normalization(group['normalization'])` in `_create_ortho_tasks`** is still redundant with the `__init__`-time canonicalization. Drop the per-step call; it's dead work in the hot path.

**Independent validation**
On the microsoft/aifsdk pipeline (1B mixformer, climbmix, FSDP2), I just enqueued a 1b/1b smoke with `MuonH(normalization='short_axis')` against this PR head as the dion pin. If it trains cleanly I'll follow up at 1b/10b for a real comparison vs the NorMuon baseline. Will report numbers if interesting.

LGTM on the fix; the remaining items are polish.

alint77:
perfect! much appreciated. please update me with the results so I can test on nanoplm too. I'm super limited on access to GPUs these days and it's super frustrating... 

I'm closely following the new Track_3 from nanogpt and trying out the new Muon varients as they pop up, will try my best to implement the ones that show real promise, so we can run at larger scales than nanogpt. 

I'm super interested in exploring some sort of second-order optimizer too... haven't really dug too deep into the literature yet but I think applying the 'bitter lesson' to the optimizers might be a promising avenue for research. especially now that we have a super throughput optimized platform/framework for rapid experimentation... 

Please let me know if anything catches your eye in the literature or you have any ideas for a more computationally heavy optimizer, I'm definitely interested in collaborating on that too.

JohnLangford:
My earlier reports were errant---it turned out I didn't have muonh on.

I did a 1b/10b scan around my current optima.  Based on that, I see: 
Neuron > Short axis (.01 nats)
Spectral norm > nothing (0 to 0.03 nats)
No muonh > spectral norm normuon (.05 nats)

I'm planning to close this for now, but welcome any new insights/ideas.

alint77:
Looks like Muse Glimmer was trained using some sort of Hyperball optimizer... My guess is to make low precision training more robust...

alint77:
forgot the link: 

https://x.com/wen_kaiyue/status/2086869377442369620

PR #75, opened 2026-05-05
megabatch: precompute padded_local_size, drop per-step allreduce sync

alint77:
The upstream empty-FSDP2-shard fix (#74) introduced an allreduce-MAX + .item() inside megabatch_orthogonalize_async to discover the rank- consistent padded local size, adding a CPU/GPU sync point on every sharded ortho task.

Compute the value locally instead: ceil(global_dim / world_size) from the still-DTensor X[0].shape in each *_update_megabatch_async caller, threaded through as a required arg when comm_dim is not None. Same correctness under FSDP2 contiguous chunking, no per-step collective.

Test: extend test_megabatch_empty_shard.py with world_size=2 cases ((1, D) empty-shard and (3, D) non-divisible) so the fast path is exercised on 2-GPU dev hosts; 4-GPU cases skip dynamically.

Before: 

Screenshot 2026-05-05 at 16 22 25

After: 

Screenshot 2026-05-05 at 16 22 44

### DIFF
--- a/dion/dion2.py
+++ b/dion/dion2.py
@@ -229,13 +229,10 @@ def dion2_update_megabatch_async(
     # comm_dim for sharded communication: use select_dim (which equals normalized shard_dim)
     comm_dim = select_dim if is_sharded else None
 
-    # Locally compute the rank-consistent padded local size from the unsharded
-    # global shape (X[0] is still a DTensor here). Avoids the per-step
-    # allreduce + .item() sync inside megabatch_orthogonalize_async.
-    if comm_dim is not None:
-        padded_local_size = -(-X[0].shape[comm_dim] // world_size)
-    else:
-        padded_local_size = None
+    # X[0] is still a DTensor here, so .shape[comm_dim] is the global
+    # (unsharded) size. The megabatch fn uses this to compute the rank-
+    # consistent pad size for its alltoall.
+    global_comm_dim_size = X[0].shape[comm_dim] if comm_dim is not None else None
 
     # Orthogonalize via shared megabatch communication
     U_ortho = yield from megabatch_orthogonalize_async(
@@ -247,7 +244,7 @@ def dion2_update_megabatch_async(
         newton_schulz_func=newton_schulz_func,
         flatten=flatten,
         epsilon=epsilon,
-        padded_local_size=padded_local_size,
+        global_comm_dim_size=global_comm_dim_size,
     )
 
     # Compute scaled learning rate
--- a/dion/dion2.py
+++ b/dion/dion2.py
@@ -229,6 +229,14 @@ def dion2_update_megabatch_async(
     # comm_dim for sharded communication: use select_dim (which equals normalized shard_dim)
     comm_dim = select_dim if is_sharded else None
 
+    # Locally compute the rank-consistent padded local size from the unsharded
+    # global shape (X[0] is still a DTensor here). Avoids the per-step
+    # allreduce + .item() sync inside megabatch_orthogonalize_async.
+    if comm_dim is not None:
+        padded_local_size = -(-X[0].shape[comm_dim] // world_size)
+    else:
+        padded_local_size = None
+
     # Orthogonalize via shared megabatch communication
     U_ortho = yield from megabatch_orthogonalize_async(
         U_selected,
@@ -239,6 +247,7 @@ def dion2_update_megabatch_async(
         newton_schulz_func=newton_schulz_func,
         flatten=flatten,
         epsilon=epsilon,
+        padded_local_size=padded_local_size,
     )
 
     # Compute scaled learning rate
--- a/dion/megabatch_base.py
+++ b/dion/megabatch_base.py
@@ -358,6 +358,7 @@ def megabatch_orthogonalize_async(
     newton_schulz_func: Callable,
     flatten: bool,
     epsilon: Tensor,
+    padded_local_size: Optional[int] = None,
 ) -> Generator[None, None, List[Tensor]]:
     """
     Shared megabatch communication + Newton-Schulz orthogonalization.
@@ -378,6 +379,10 @@ def megabatch_orthogonalize_async(
         newton_schulz_func: Newton-Schulz orthogonalization function.
         flatten: Whether to flatten 3D+ tensors to 2D.
         epsilon: Small value for numerical stability.
+        padded_local_size: Required when ``comm_dim is not None``. The
+            rank-consistent padded local size along ``comm_dim``, computed by
+            the caller as ``ceil(global_dim / world_size)`` from the unsharded
+            DTensor's global shape. Avoids a per-step allreduce + GPU->CPU sync.
     """
     N = len(U)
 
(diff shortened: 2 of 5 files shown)


JohnLangford:
This is John's reviewbot.

## Review: megabatch: precompute padded_local_size, drop per-step allreduce sync

**Summary:** Solid fix. Eliminating the allreduce+.item() sync is the right call — that's a hard GPU stall on every optimizer step. The core math is correct and the assert is good defensive programming. A few issues worth addressing before merge.

---

### Correctness

**Core computation is correct.** `-(-X[0].shape[comm_dim] // world_size)` is ceiling division and produces the same value as `allreduce-MAX(local_size)` under FSDP2's contiguous chunking. The comment correctly documents why.

**Critical assumption: FSDP2 contiguous chunking.** The PR assumes `padded_local_size = ceil(global_dim / world_size)`, which holds because FSDP2 assigns contiguous chunks. But this is a behavioral property of the current FSDP2 implementation, not a contractual guarantee. If FSDP2 ever changes its sharding strategy (e.g., block-cyclic), `padded_local_size` could be wrong and `all_to_all` would silently hang or corrupt data — the same silent failure mode the original bug introduced. Worth adding a comment explicitly calling out this dependency, and potentially adding a debug-mode assert that `padded_local_size >= U_work[0].size(comm_dim)` inside `megabatch_orthogonalize_async` (cheap, catches the failure early).

**`world_size` coupling.** The computation in each caller uses the `world_size` parameter (passed in from outside), while `process_group` is passed separately into `megabatch_orthogonalize_async`. These must be consistent. There's no enforcement of this. If a caller ever passes the wrong `world_size` (e.g., global world size instead of the process group's size), the padding will be wrong. Consider computing `world_size` from `dist.get_world_size(process_group)` inside `megabatch_orthogonalize_async` instead of relying on the caller to pass the right value.

---

### Design

**Code duplication.** The same 6-line block appears identically in `dion2.py`, `muon.py`, and `normuon.py`. Extract it:

```python
def _compute_padded_local_size(x: Tensor, comm_dim: Optional[int], world_size: int) -> Optional[int]:
    if comm_dim is None:
        return None
    return -(-x.shape[comm_dim] // world_size)
```

**API awkwardness.** `padded_local_size: Optional[int] = None` with an assert that fires when it's `None` and `comm_dim is not None` is a soft required argument. If this is internal API, just make the callers always pass it. Alternatively, since `process_group` is already available inside `megabatch_orthogonalize_async` and the global dim could be passed as an int, you could keep the computation inside the function and pass `global_comm_dim_size: Optional[int]` instead — cleaner separation of what the caller knows vs. what the function needs.

**More elegant alternative.** Instead of threading `padded_local_size` through the call chain, pass the global dimension directly:

```python
# In megabatch_orthogonalize_async signature:
global_comm_dim: Optional[int] = None  # X[0].shape[comm_dim] from DTensor global shape

# Inside the function, when comm_dim is not None:
world_size = dist.get_world_size(process_group)
padded_local_size = -(-global_comm_dim // world_size)
```

This consolidates both the responsibility and the `world_size` source, avoids the three-site duplication, and makes the function self-contained given only information the caller has.

---

### Test Coverage

**Good additions.** The 2-GPU parametrize cases `(1, 2, 8)` and `(3, 2, 8)` cover the empty-shard and non-divisible paths without requiring 4 GPUs. The dynamic skip is strictly better than the class-level decorator.

**Gap: callers are not tested.** The new test exercises `megabatch_orthogonalize_async` directly. There are no tests for the `padded_local_size` computation in `dion2_update_megabatch_async`, `muon_update_megabatch_async`, `normuon_update_megabatch_async`. If a caller passes the wrong `world_size`, the test would pass while the real code breaks. A light integration test that exercises one of the `*_update_megabatch_async` entry points with an uneven shard would close this gap.

---

### Performance

The change is unambiguously correct on the performance dimension — one fewer collective + H2D sync per optimizer step is a win on every FSDP2 run. The screenshots confirm the sync is gone.

---

### Backward Compatibility

`padded_local_size` has a default of `None` so call sites that omit it still compile. The assert gates correctness at runtime when the sharded path is actually hit. Any external caller using `megabatch_orthogonalize_async` with `comm_dim is not None` must now pass `padded_local_size` or they get an `AssertionError` at step 0 — acceptable for what appears to be an internal API, but worth documenting in the changelog if this is a public-facing library.

---

### Verdict

Approve with suggestions. The duplication across three files and the FSDP2-assumption comment are the most important things to address. The `world_size` coupling is lower priority but worth fixing to make the invariant explicit.

alint77:
@JohnLangford fixed the code duplication issue by passing global_comm_dim, let me know if there's anything else needs changing.

JohnLangford:
This is John's reviewbot.

**Verdict:** correct, well-justified, and a real win. Removing the per-step `all_reduce(MAX) + .item()` kills a CPU/GPU sync that was firing on every sharded ortho task, which matters a lot when the rest of the path is async. Comments and tests both improved. A few thoughts below.

### Correctness
- The derivation `padded_local_size = ceil(global / world_size)` matches FSDP2's `_chunk_with_empty(t, W, dim=shard_dim)` exactly: chunk size is `ceil(D/W)`, ranks beyond `floor(D / ceil(D/W))` get size 0, and rank 0 always holds the max. So this number equals what the old `all_reduce(MAX)` produced. ✓
- The `assert padded_local_size >= original_local_size` is a good belt-and-braces invariant catching any future FSDP sharding-strategy change.
- The new value can never be *smaller* than the true max under FSDP2 contiguous chunking, and being equal-or-larger only ever adds zero rows that Newton-Schulz preserves, so even a hypothetical overestimate would be correct (just wasteful).

### Suggestions
1. **`assert` is the wrong tool for an external-contract check.** `assert global_comm_dim_size is not None` and the size-assertion both vanish under `python -O`. If anyone runs with optimizations enabled, a missing arg silently becomes `TypeError: unsupported operand type(s) for -: 'int' and 'NoneType'` deep inside the formula, or worse a wrong `padded_local_size`. Prefer `if global_comm_dim_size is None: raise ValueError(...)`. Same for the sharding-invariant check — that one is genuinely defensive, but a `RuntimeError` survives `-O`.
2. **Type signature lies a little.** `global_comm_dim_size: Optional[int] = None` says "optional", but it's actually required when `comm_dim is not None`. Either drop the default (callers in this PR all pass it) or split the API. Minor — the docstring does call it out.
3. **`-(-x // y)` works but `(x + y - 1) // y` reads more obviously as ceiling division** to anyone not steeped in the idiom. Taste; not blocking.
4. **Alternative factoring worth considering:** the megabatch fn doesn't actually need to know about FSDP2 at all if callers pass `padded_local_size` directly. Each caller already has `global_dim`, `world_size`, and the FSDP2 assumption — they could compute `ceil` themselves and hand the function a pre-computed pad target. That keeps `megabatch_base.py` agnostic of sharding strategy and pushes the FSDP2-specific NOTE comment to the call sites that actually depend on FSDP2. Current factoring is fine; this would just be cleaner separation of concerns.

### Tests
- Adding `world_size=2` cases is the right call — `(1, D)` covers the empty-shard branch on a 2-GPU dev box, `(3, D)` covers non-divisible. Good.
- Replacing the class-level skip with per-parametrization `pytest.skip` is strictly better.
- One gap: no test exercises the dion2/muon/normuon caller path with an actual DTensor. The unit test passes `global_comm_dim_size=global_dim_0` directly, so it doesn't validate the "X[0] is still a DTensor here, so `.shape[comm_dim]` is global" claim in the three call sites. A small integration test (or even an assertion `isinstance(X[0], DTensor)` at extraction) would catch a future refactor that accidentally moves `to_local(X)` above the extraction.

### Backward compat / security / perf
- Backward compat: signature change is additive with a default, so external callers compile; sharded external callers will hit the new assert. Acceptable for an internal-ish API.
- Security: n/a.
- Perf/throughput: claim is consistent with the code change — one fewer collective and one fewer device→host sync per sharded ortho task per step. The screenshots back it up.

Nice cleanup.

Issue #76, opened 2026-05-06
NaN param values introduced for multi gpu training between 3a52bdf and e8e1c6a


skaae:
Hi, 

First thank you for great library!

After updating from commit 3a52bdf  to e8e1c6a we started having training jobs fail where the parameters will be NaN. I have indentified that the gradients are ok, but parameters will be Nan after optimizer.step has been called. We train using 6 blackwell 6000 gpus using DDP and nccl. 

We have 6 blocks in our network. the failing layer is always the linear layer with shape [512, 1408] in one of the blocks. Its not deterministic what block that fails, but its always the [512, 1408]  that fails. Its also not deterministic when the error happens, but usually after 0-30 min of training. 

I have so far confirmed that the gradients are fine until `optimizer.step()` is called. After the optimizer.step() the layer weight with shape [512, 1408]  will be all Nans. I therefore suspect its related to the update the of the Newton-Shultz package or maybe Quack Kernels being bumped? Alternatively the megabatch PR. 

Sorry for the not super specific bug report.
I'm happy to dump some better debug data, but I would need some guidance on what to dump. 

```
Driver Version: 590.48.01      CUDA Version: 13.1  
torch                              2.11.0+cu130


OK version:
dion commit 3a52bdf
quack kernels pinned to 0.3.10

Failing version:
dion commit e8e1c6a
```

alint77:
Have you tested PR #75 ?

JohnLangford:
This is super-important to nail, but I need some help because I can't reproduce.
--Have you tried a bisection search for the exact commit?
--what are the flags you are using exactly?
--Have you tried swapping the optimizer cores (gram ns with/without triton vs. polar express)?

skaae:
I Spent yesterday narrowing it down to the optimizer step.  



**--Have you tried swapping the optimizer cores (gram ns with/without triton vs. polar express)?**
No. Do you have some combinations you want me to try? 

I can also just dump the input/ouput updates of the normuon internals? I'm not super familiar with NorMuon but if you can point me to the most likely place where this could happen inside NorMuon, then I can check for nans and save the input that resulted in Nans. 

From looking at the code I was about to check for nans around this function: 

```
    U = yield from megabatch_orthogonalize_async(
        U,
        comm_dim=comm_dim,
        device_rank=device_rank,
        world_size=world_size,
        process_group=process_group,
        newton_schulz_func=newton_schulz_func,
        flatten=flatten,
        epsilon=epsilon,
    )
```

[...]

PR #78, opened 2026-05-08
megabatch: env-gated NaN capture wrapper around newton_schulz_func

JohnLangford:
## Summary

Re: #76. The reporter and maintainers cannot consistently reproduce the
NorMuon + gram-newton-schulz + quack-kernels 0.4.1 NaN bug, so the next
debugging step is to capture the actual NS input that triggers the failure
on the affected hardware (6x Blackwell RTX PRO 6000 + DDP).

This PR wraps ``newton_schulz_func`` inside ``megabatch_orthogonalize_async``
with an env-gated capture wrapper. When enabled, each rank that observes a
non-finite NS output writes its own .pt dump (input + output + rank +
shape + epsilon) and by default raises ``RuntimeError``. The check is
**local per rank — no extra collective** — and it short-circuits at one
``os.environ`` lookup when disabled.

## Usage

```bash
DION_NAN_CAPTURE=1 \
DION_NAN_CAPTURE_DIR=/some/scratch/dir \
DION_NAN_CAPTURE_RAISE=1 \  # default; set to 0 to keep training
torchrun ... train.py
```

Per-rank filenames are ``dion_nan_capture_rank{rank}_shape{...}_pid{...}_{ts_ms}.pt``
so concurrent writes from multiple ranks on a shared filesystem do not collide.

## Notes

- Disabled by default. Cost when off: one ``os.environ`` lookup per call.
- The input is ``.detach().clone()``-d up front so we still have the
  offending tensor even if a kernel mutates X in-place.
- Two new module-level imports (``os``, ``time``).

## Test plan

- [x] ``test_capture_disabled_by_default`` — no env, no dumps even when NS returns NaN
- [x] ``test_capture_dumps_on_non_finite_output`` — dump contains correct input/output/shape/rank/epsilon
- [x] ``test_capture_raises_by_default_when_enabled`` — ``RuntimeError`` after dump
- [x] ``test_capture_no_dump_when_finite`` — finite output -> no dump even with env on
- [x] ``test_capture_filename_includes_rank`` — non-zero rank reflected in filename

All five pass single-rank without GPU/NCCL.

### DIFF
--- a/dion/megabatch_base.py
+++ b/dion/megabatch_base.py
@@ -1,4 +1,6 @@
 import math
+import os
+import time
 import torch
 import torch.distributed as dist
 from collections import defaultdict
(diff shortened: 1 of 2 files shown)


donglixp:
https://github.com/microsoft/dion/pull/102
This PR might be related to the above issue.

JohnLangford:
Closing for now---it may be addressed.

PR #79, opened 2026-05-08
NorMuon: opt-in nan_guard_fallback skips step on non-finite NS output

JohnLangford:
## Summary

Re: #76. Defensive guard for the intermittent NaN bug in NorMuon +
gram-newton-schulz + quack-kernels 0.4.1: once a single rank's NS output
goes non-finite, the bad value poisons the parameter, all-reduces into
the next step, and the run is dead. With this PR, opting into
``nan_guard_fallback=True`` lets all ranks agree (via a single
``all_reduce(MAX)`` of one byte) to skip the entire post-ortho update on
detection.

- ``V`` (variance buffer) and ``X`` (param) stay strictly unchanged on a
  skipped step. ``M`` (momentum) was updated from the clean gradient in
  pre-orthogonalize and is left alone.
- The early-return semantics are deliberately "this batch never
  happened" rather than "zero out U and run normalization", because
  decaying V on zero updates contaminates future steps and applying
  weight decay alone is a half-update we don't want.

Companion to #78 (the env-gated capture wrapper that helps identify the
offending input). #78 is for diagnosis; this PR is for surviving the
issue in production until the upstream quack/gns regression is fixed.

## Cost

| | Off (default) | On |
|---|---|---|
| Per-step | 1 Python branch | 1-byte ``all_reduce(MAX)`` + ``.item()`` per shape group |
| Behavior change | None | Skip step on detection; emit RuntimeWarning on rank 0 |

The sync is required: if rank 0 takes the fallback and rank 1 doesn't,
DDP weights drift or the next megabatch alltoall mismatches and
deadlocks. Single-byte allreduce is microseconds; total overhead is
dominated by the device->host ``.item()`` sync.

## Usage

```python
opt = NorMuon(
    params,
    distributed_mesh=process_group,
    use_gram_newton_schulz=True,
    use_triton=True,
    nan_guard_fallback=True,  # opt-in defense for issue #76
)
```

## Test plan

- [x] ``test_fallback_off_lets_nan_propagate_to_params`` — baselines that
      the bug actually exists in unit-test form so the positive test isn't
      trivially green
- [x] ``test_fallback_on_skips_step_when_ns_returns_nan`` — bit-exact
      param + zero variance buffer + RuntimeWarning on rank 0
- [x] ``test_fallback_on_does_not_block_normal_step`` — guard is inert
      when NS output is finite
- [x] Existing ``test_optimizers.py::TestNorMuon`` (5 tests) still pass

All run single-rank on CUDA without NCCL.

### DIFF
--- a/dion/normuon.py
+++ b/dion/normuon.py
@@ -167,6 +167,13 @@ class NorMuon(DistributedOrthoBase):
                 newton_schulz_func=self._newton_schulz_func,
                 cautious_wd=group["cautious_wd"],
                 nan_guard_fallback=self._nan_guard_fallback,
+                # Sync the nan flag on the optimizer's full process group,
+                # not the megabatch's: the head-split and batch-sharded
+                # paths below override ``process_group`` to ``None`` (no
+                # alltoall is needed there), but the nan-skip decision
+                # still has to agree across all ranks or DDP drifts and
+                # sharded params end up torn.
+                nan_sync_process_group=self._process_group,
             )
 
             shape_groups: dict[tuple, list] = defaultdict(list)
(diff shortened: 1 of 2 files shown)


JohnLangford:
Likely unnecessary, closing for now.

PR #80, opened 2026-05-10
Add Aurora optimizer for non-square matrices

JohnLangford:
## Summary
- Adds Aurora ([blog](https://blog.tilderesearch.com/blog/aurora), [reference impl](https://github.com/tilde-research/aurora-release)) as a new optimizer in `dion/aurora.py`, exported as `dion.Aurora`.
- Aurora is a leverage-uniform polar update for non-square matrices: it iteratively row-preconditions the gradient before each polar (Newton-Schulz) call so that all left-singular rows of the orthogonalized update have comparable norms. For square matrices it reduces to standard Muon polar bit-for-bit.
- Reuses Muon's pre/post-orthogonalize stages and the shared megabatch infrastructure; the new behavior lives entirely in a `newton_schulz_func` wrapper, so mega-batching, FSDP2 sharding, `num_heads`, and mixed param groups all work transparently.

## What changed
- `dion/aurora.py` (new): `Aurora` class + `make_aurora_polar(base_polar, pp_iterations, pp_beta)` factory. Default hyperparameters match the reference (`pp_iterations=2`, `pp_beta=0.5`, `nesterov=True`, `adjust_lr=None`).
- `dion/__init__.py`: export `Aurora`.
- `tests/test_optimizers.py`: new `TestAurora` class (11 tests) covering basic operation, determinism, parameter update, Nesterov, cautious WD, `pp_iterations` sweep, megabatching, mixed shapes, validation. Two Aurora-specific assertions:
  - **Square matrices**: Aurora's polar output is bit-for-bit identical to the standard polar.
  - **Non-square matrices**: Aurora's row-norm `max/min` ratio is meaningfully tighter than standard polar (and below 1.15).

## Algorithm
For non-square `G` with `(m, n)`:
1. Transpose to tall (`m >= n`).
2. Initialize `D = 1 / row_norm(G)`.
3. For `pp_iterations` rounds: `U = polar(D * G)`; between rounds update `D *= (target_row_sq / row_sq(U))^pp_beta` where `target_row_sq = n / m`.
4. Transpose back if needed; multiply by `max(1, m/n)^0.5` (Aurora's tall-aspect-ratio scaling, baked into the orthogonalization output rather than the LR).

For square matrices the loop is bypassed entirely.

## Test plan
- [x] `pytest tests/test_optimizers.py::TestAurora` — 11/11 pass on CUDA.
- [x] `pytest tests/test_optimizers.py tests/test_polar_express.py tests/test_newton_shulz.py` — 92/92 pass (no regressions).
- [x] Verified empirically on a tall (512×128) matrix: Aurora row-norm `std` drops 6× (0.029 → 0.005) and `max/min` ratio improves from 1.40 to 1.06.
- [x] Verified `Aurora == polar` bit-for-bit on square (128×128) input.
- [ ] FSDP2 / DTensor sharded smoke test (not run; the wrapper plugs into the existing megabatch path that NorMuon also uses, so coverage is via the shared infrastructure).

### DIFF
--- a/dion/aurora.py
+++ b/dion/aurora.py
@@ -1,0 +1,369 @@
+import torch
+from collections import defaultdict
+from torch import Tensor
+from torch.distributed import ProcessGroup
+from torch.distributed.tensor import DeviceMesh, DTensor
+from torch.optim.optimizer import ParamsT
+from typing import Callable, Generator, List, Optional, Tuple, Union
+
+from .megabatch_base import (
+    DistributedOrthoBase,
+    megabatch_orthogonalize_async,
+    adjust_lr_spectral_norm,
+    adjust_lr_rms_norm,
+)
+from .muon import muon_update_pre_orthogonalize, muon_update_post_orthogonalize
+from .newton_schulz_triton import (
+    TRITON_AVAILABLE,
+    newton_schulz_triton,
+    zeropower_via_newtonschulz5,
+)
+from .opt_utils import AsyncTask, to_local
+from .polar_express import polar_express, polar_express_triton
+
+
+class Aurora(DistributedOrthoBase):
+    """
+    Distributed Aurora optimizer for PyTorch FSDP2. Also compatible with DDP.
+
+    Aurora is an optimizer for non-square weight matrices that achieves more
+    balanced neuron utilization than standard Muon. Instead of applying the
+    polar (Newton-Schulz) factor directly, which inherits non-uniform
+    left-singular row norms, Aurora iteratively approximates a projection onto
+    the intersection of the row oblique and Stiefel manifolds via diagonal
+    preconditioning. The result is a leverage-uniform update.
+
+    For square matrices Aurora reduces to standard Muon.
+
+    Args:
+        params: Parameters for the optimizer.
+        distributed_mesh: DeviceMesh or ProcessGroup for distributed training.
+            Use DeviceMesh for FSDP2 and ProcessGroup for DistributedDataParallel.
+        lr: Base learning rate. Aurora bakes the tall-aspect-ratio scaling
+            ``max(1, m/n)**0.5`` into the update, so ``adjust_lr`` defaults to None.
+        mu: Momentum factor.
+        betas: Tuple of (beta1, beta2) for AdamW and Lion algorithms.
+        weight_decay: Weight decay factor.
+        cautious_wd: Whether to apply weight decay only where update and parameter signs align.
+        epsilon: Small value to avoid division by zero.
+        nesterov: Whether to use Nesterov momentum.
+        adjust_lr: Optional Muon-style LR adjustment ("spectral_norm", "rms_norm", or None).
+            None is the Aurora default; the algorithm already applies its own
+            ``max(1, m/n)**0.5`` aspect-ratio scaling inside the orthogonalization.
+        flatten: Whether to flatten 3D+ tensors to 2D for the orthogonalization step.
+        pp_iterations: Number of preconditioned-polar iterations. Each iteration
+            calls the base polar (Newton-Schulz) once. ``pp_iterations=2`` is the
+            Aurora paper default; ``pp_iterations=1`` is single-shot row-norm
+            preconditioning.
+        pp_beta: Exponent for the diagonal update between iterations.
+        use_triton: Whether to use the Triton Newton-Schulz kernel.
+        use_polar_express: Whether to use Polar Express for the base polar.
+        newton_schulz_func: Optional custom base polar function. Aurora wraps
+            this with its diagonal preconditioning loop.
+
+    Aurora: https://blog.tilderesearch.com/blog/aurora
+    Reference implementation: https://github.com/tilde-research/aurora-release
+    """
+
+    def __init__(
+        self,
+        params: ParamsT,
+        distributed_mesh: Optional[Union[DeviceMesh, ProcessGroup]] = None,
+        lr: float = 0.01,
+        mu: float = 0.95,
+        betas: Tuple[float, float] = (0.9, 0.95),
+        weight_decay: float = 0.01,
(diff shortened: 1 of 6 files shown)


dion/aurora.py
```diff
@@ -0,0 +1,390 @@
...
+        if not isinstance(pp_iterations, int) or pp_iterations < 1:
+            raise ValueError(
+                f"Invalid pp_iterations: {pp_iterations}. Must be a positive integer."
+            )
+        if pp_beta < 0.0:
+            raise ValueError(f"Invalid pp_beta: {pp_beta}")
+
+        # Resolve the base polar function (the one wrapped by Aurora's
+        # diagonal preconditioning). Mirrors DistributedOrthoBase resolution.
+        if newton_schulz_func is not None:
```
NoahAmsel, line 115: arora's polar function uses 12 iterations. maybe too soon to decide, but should we add an option to use their polar function, or at least to do more than 5 iterations? (it will take a little care to integrate that with gram newton schulz, but it's doable)

dion/aurora.py
```diff
@@ -0,0 +1,390 @@
...
+    adjust_lr: Optional[str],
+    device_rank: int,
+    world_size: int,
+    shard_dim: Optional[int] = None,
+    process_group: Optional[ProcessGroup] = None,
+    newton_schulz_func: Optional[Callable] = None,
+    cautious_wd: bool = False,
+) -> Generator[None, None, None]:
+    """
+    Mega-batched Aurora update. Reuses Muon's pre/post-orthogonalize stages
```
NoahAmsel, line 287: So what is the difference between this function and the generic update_megabatch_async?

tests/test_optimizers.py
```diff
@@ -93,6 +93,141 @@ def test_mixed_shapes(self):
...
+        assert torch.equal(r1[0], r2[0])
+
+    def test_params_change(self):
+        from dion import Aurora
+        params = _make_params([(64, 128)])
+        before = params[0].data.clone()
+        _run_steps(Aurora, params, dict(lr=0.01), n_steps=1)
+        assert not torch.equal(params[0].data, before)
+
+    def test_nesterov(self):
```
NoahAmsel, line 122: what does this test? just that there's no error? same for `test_cautious_wd` and `test_pp_iterations`

NoahAmsel (approved):

NoahAmsel:
Questions
1. Arora is concerned about the up projections (tall/skinny) only, since that's where they observe the neuron death phenomenon. Why do they even apply all this extra normalization to short / fat matrices? Why should we?
2. I don't really understand why they design the damping parameter as they do. It does however seem unnecessary to multiply D by target_row_sq, which is constant across rows, since polar is invariant to scaling by a constant anyway. Probably no harm though

JohnLangford:
The first comment seems valid in my experiments---I'm seeing the best results with spectral norm (no min).  So far, it seems like the ideal learning rate is more like 0.01, but there's evidence that this provides a modest advantage over normuon.

JohnLangford:
Tests at 1b/100b with learning rate 0.01 and spectral norm reveal perf somewhat worse than normuon.   In addition Noah analyzed things at 1b/1b scale and can't replicate observations of 'neuron death'.   For unclear reasons, 
(a) it seems that aurora is addressing a problem not evident in our training models.  
(b) it seems that we can't use the very high learning rate (0.05) discussed in the paper.

alint77:
I tested on a 170M proteinLM with srelu 3x expansion factor, applying aurora to the mlp only, both the loss and downstream task performance were better comapared to normuon

JohnLangford:
Interesting.   I wish I understood what was the important difference here w.r.t. your setup.

aldew5:
Hey all I'm one of the authors on Aurora. Stumbled across this thread (a little late) and thought the discussion was interesting so wanted to add my perspective (I've also found the dion repo to be very helpful in general).

@NoahAmsel It's true that Aurora is only intended for updates to tall matrices. The release code was confusing for a couple reasons: the MLP only had wide matrices so it really didn't make any sense as a demo model, and the NS routine was simple quintic which tends to be much less precise than Polar Express or CANS. Part of the reason we only apply Aurora on tall matrices is the rows of a semi-orthogonal wide matrix are already uniform, but simple quintic can fail to approximate this well. I've updated the repo to use a SwiGLU MLP (which actually has tall params) and the reference implementation now only updates tall matrices, which is recommended. 

The goal of the damping was to reduce overshoot and oscillation due to row coupling. We found it made the iteration converge better but there is plausibly a better design. Good point about the pre-polar scaling as well. The thought was that allowing the scale of D to drift could cause problems in edge cases, so we included the rescaling as a cheap safety measure. 

@JohnLangford it's true that the extent of underutilized/dead neurons has some dependence on training setup that I don't fully understand yet (but working on this). Although I think this problem is fairly general. I had some results here (https://arxiv.org/abs/2606.27715) showing dead neuron phenomena in open-weight models (but less extreme than what we found in the blog). I've also found Aurora's row normalization can be useful even when neurons aren't fully dead but underutilized in a less dramatic way. This may offer some explanation on why Aurora can be useful in settings where Muon is not yielding dramatically dead neurons. 

Other notes:
- Lr of 0.05 is probably too big. I think this only worked for nano-GPT. For pre-training experiments, I've been using lr on the scale of 1e-4.
- Helpful to use precise NS like Polar Express or CANS
- Not sure the cases in which NorMuon may do better but haven't observed this in my settings. I also find that Aurora tends to scale pretty well but NorMuon hasn't performed as well when I've tested on slightly bigger models (~4B-8B).

JohnLangford:
Thanks @aldew5 , very helpful.  I'm currently paused on experiments here, but we may get back to it.

JohnLangford:
Closing this for now---I haven't yet found a good use case in my experiments.   I'm happy to reconsider with a good example of where it's useful.

PR #82, opened 2026-05-27
Add NorDion2 optimizer combining NorMuon and Dion2 paths

ayfeng23:
## Summary

Implement NorDion2 optimizer to reduce orthonormalization computation through Dion2 row-wise selection while preserving NorMuon-style updates. 

## Changes

- **`dion/nordion2.py`** (new): Adds `NorDion2` class which updates parameters via `dion2_pre_orthogonalize`, megabatch orthogonalization, stacked normuon normalization, and `dion2_post_orthogonalize`.
- **`tests/test_optimizers.py`**: Added 7 unit tests covering basic correctness across parameter shapes, determinism, and parameter updates. `NorDion2`-specific tests verify fraction sweep, dtype preservation, `variance_neuron` shape, and megabatch grouping.
- **`dion/__init__.py`**: Export `NorDion2` from package.
- **`train.py`**: Add `nordion2` optimizer option to training script.

## Algorithm and Design Choices
Designed to rely on Dion2 and NorMuon functions so that further optimizations transfer to NorDion2. As NorMuon's `variance_neuron` tracks one parameter per row, we restrict NorDion2 to row-wise selection so that selected rows update their respective variance estimates. 

1. Use `dion2_pre_orthogonalize` to add gradient to momentum buffer and select top `fraction` rows for orthogonalization. Decays momentum buffer afterward. 
2. Orthonormalizes selected rows into `U_ortho` and select variance buffer for corresponding rows to get `V_sel`.
3. Upcast `V_sel` (and `U_ortho` indirectly) to fp32 computation.
4. Normalize `U_ortho` using `normuon_normalization_stacked` and then update parameters.

NorDion2 at fraction=1.0 (1-NorDion2) differs from NorMuon in four ways that affect performance minimally:
1. Top `fraction` row selection permutes rows
2. Momentum damping is applied after, so that `M + G` is orthonormalized, and not `momentum * M + G`, where `momentum` is the damping factor
3. fp32 normalization is used for higher precision to reduce quantization error. At fraction < 1, each row's variance is updated less frequently, leading to fewer corrections to quantization error 
4. post_orthogonalization uses Dion2's implementation and not Muon's

## Testing
- Evaluations on eval/0/loss for 1b models with triton_post_ortho=True
  - 10b tokens: 1-NorDion2 3.0685, NorMuon 3.0632
  - 30b tokens: 1-NorDion2 2.9438, NorMuon 2.9405
- Ablations applying the above 4 changes to NorMuon for 10b tokens
  - (1) 3.0651, (2) 3.0635, (3) 3.0634, (4) 3.0640
- All unit tests passed


ayfeng23:
@microsoft-github-policy-service agree company="Microsoft"

JohnLangford:
This is John's reviewbot.

Nice, focused composition — reusing `dion2_pre_orthogonalize`/`dion2_post_orthogonalize` + `normuon_normalization_stacked` is the right instinct, the last-dim-shard guard mirrors NorMuon, and doing the variance EMA in fp32 (vs NorMuon's bf16) is a genuine precision improvement. The eval numbers look like clean parity with NorMuon. A few substantive issues:

## 1. Blocking correctness: crashes on any parameter with `ndim >= 3` (native 3D and `num_heads`)

The new variance handling uses

```python
selected_v = v.index_select(dim=select_dim, index=indices)   # select_dim = -2
...
v.index_copy_(dim=select_dim, index=indices, source=...)
```

`index_select`/`index_copy_` require a **1-D** index vector. But `dion2_pre_orthogonalize` returns *batched* top-k indices: for a 3-D input (native batched matrix, or a 2-D weight reshaped by `_prepare_head_split` into `(heads, head_dim, cols)`), each entry of `indices_list` has shape `(heads, k)` / `(batch, k)`, not `(k,)`. Reproduced on this branch (CUDA):

```
Case 1: native 3D param (4, 8, 16)      -> RuntimeError: Index is supposed to be ... a vector
Case 2: 2D param (32,16) num_heads=4    -> RuntimeError: Index is supposed to be ... a vector
Baseline: 2D param (64,128)             -> OK
```

This isn't a theoretical path: `_create_ortho_tasks` explicitly calls `_prepare_head_split(num_heads, params, gradients, momentums, variances_neuron)` and the assert only requires `p.ndim >= 2`, so NorDion2 *advertises* head/3D support it doesn't have. Dion2 avoids this by using `gather`/`scatter_add_` with expanded indices (which handle batched indices); NorMuon avoids it by never selecting `V`. Fix options:

- Mirror Dion2's `gather`/`scatter`-with-`expand` pattern for the `V` select and write-back, **or**
- If 3D/head support is out of scope, gate it with a clear `NotImplementedError` (as `_get_shard_info` already does for last-dim shards) and stop threading `variances_neuron` through `_prepare_head_split`.

## 2. Test coverage gaps (this is why #1 slipped through)

- **No `test_nordion2_matches_3d`** in `TestNumHeads`. Muon, Dion2, and NorMuon each have a 2D-vs-3D parity test there; adding the NorDion2 analogue would have caught #1 immediately.
- **The deployed path is untested.** All NorDion2 tests use the default `triton_post_ortho=False` (eager), but `train.py` and the reported 1B runs use `triton_post_ortho=True`. The Triton post-ortho branch has zero unit coverage.
- **No sharded/distributed coverage** — the `comm_dim`/all-to-all/DTensor path and the new last-dim-shard `NotImplementedError` are never exercised (the suite looks single-GPU generally, so this may be a known limitation, but worth noting given NorDion2's whole point is the sharded selection).
- `test_fraction` iterates `[0.1, 0.25, 0.5, 1.0, 1]` — `1.0` and `1` are duplicates. Consider asserting that `fraction=1.0` selects all rows and updates every `variance_neuron` entry.

## 3. Design: ~250 lines duplicated from NorMuon

`_create_ortho_tasks`, `_get_shard_info`, `_get_or_initialize_state`, and most of `nordion2_update_megabatch_async` (the shape-grouping, head-split, shard-detection, `global_comm_dim_size` block, and lr-adjust dispatch — comments included) are near-verbatim copies of NorMuon's, differing only in (a) Dion2 pre/post-ortho swapped in and (b) `V` row-selection. The stated goal — "rely on shared functions so optimizations transfer to NorDion2" — is undercut by re-implementing the orchestration a third time: any future fix to the megabatch/shard scaffolding (or to bug #1's index handling) now has to land in three files. A more elegant structure would factor the per-algorithm `pre_ortho -> normalize -> post_ortho` steps into hooks/strategy methods on `DistributedOrthoBase` (which already centralizes `step`, shard detection, and the scalar-opt tasks), so NorDion2 becomes "Dion2 selection + NorMuon normalize" composed rather than copied.

## 4. Minor

- `import math` in `nordion2.py` is unused.
- `dion/__init__.py` is missing a trailing newline.
- `muon_beta2` is hardcoded to `0.95` both as the class default and in `train.py`; fine (matches NorMuon) but not surfaced as an HP.

## Other dimensions
- **Performance/throughput:** the fp32 upcast is only over selected rows, so the overhead is minor; orthonormalizing `k = fraction*rows` rows instead of the full matrix delivers the intended NS savings. One note: forcing `select_dim = -2` (required by the per-row variance) means NorDion2 can't pick the shorter dimension the way Dion2 does, so for `rows >> cols` matrices it orthonormalizes a taller submatrix than Dion2 would — probably fine for typical transformer shapes, worth a comment.
- **Backward compatibility:** clean — purely additive (new file, new export, new `train.py` branch); no existing optimizer touched.
- **Security:** nothing of concern.

Happy to re-review once #1 has a fix + parity test.

JohnLangford:
This is John's reviewbot.

Re-reviewed commit `989c35c7` ("Address John reviewbot comments"). The blocking issue is fixed and the fixes check out — verified locally (CUDA):

- **3D / `num_heads` crash → fixed.** Swapping `index_select`/`index_copy_` for `gather`(dim=-2, expanded index) / `scatter_` is the right move and matches Dion2's batched-index handling. My two original repro cases now pass (native 3D `(4,8,16)` and 2D `(32,16)`+`num_heads=4`), and the shape logic is correct for 2D `(N,k,1)`, native-3D `(N,batch,k,1)`, and the head-split view.
- **Coverage added.** `test_nordion2_matches_3d` / `_full_fraction` (these would have caught the original bug), `test_triton_post_ortho` + `_parity` (the actually-deployed path), and the `fraction=1.0` variance-update assertion. All 11 NorDion2 tests pass for me (`pytest -k NorDion2`, 25s).
- Unused `import math` and the missing `__init__.py` newline are gone. 👍

Two items from the prior review remain, both **non-blocking**:

1. **Design / duplication (unaddressed).** `_create_ortho_tasks` + `nordion2_update_megabatch_async` are still ~250 lines near-verbatim from NorMuon. This bug is a good illustration of the cost: the identical index-selection logic had to be reasoned about fresh here instead of inherited. Factoring per-algorithm `pre_ortho → normalize → post_ortho` into hooks on `DistributedOrthoBase` would make NorDion2 "Dion2 selection + NorMuon normalize" by composition. Fine to defer, but worth a tracking note.
2. **No sharded/distributed test.** NorDion2's raison d'être is the sharded row-selection (`comm_dim`/all-to-all/DTensor) path, which is still only exercised single-GPU. Appears to be a suite-wide gap, not specific to this PR, but flagging since the new `gather`/`scatter_` runs on `to_local` shards and the multi-rank index/`global_comm_dim_size` interaction is untested.

Nit: `test_fraction`'s `[..., 1.0, 1]` still has the duplicate `1.0`/`1`.

LGTM once you decide whether to track (1). Nice turnaround.

ayfeng23:
# Reply to John's reviewbot
## 3D and `num_heads` support

Fixed the variance handling to use Dion2's gather / scatter approach. `indices` needs to be reconstructed by stacking `indices_list.` While we could integrate the `V` selection logic into the pre_orthogonalize function to avoid this, I chose to preserve the current code so we can utilize `dion2_pre_orthogonalize`; there should be minimal overhead.

## Test coverage

1. Fixed, added the 2D-vs-3D parity tests for fractions 0.5 and 1.0
2. Added tests to check for runtime failure and parity, as runs with `triton_post_ortho=True/False` should perform similarly
3. Existing test suite does not include similar tests, so this may be out of scope. This would be difficult to test as it requires multi-GPU setup
4. Kept duplicate to match `Dion2`. Added to `test_fraction` to ensure that all `variance_neuron` entry updates after one step at `fraction=1.0`

## Design
I decided to keep the existing design. My implementation of `NorDion2` seeks to mimic what `Dion2` does for `Muon`; `create_ortho_tasks` and `*_update_megabatch_async` are implemented across `Dion2`, `Muon`, and other optimizers. As such, I believe it is appropriate to maintain the design. The added code from `_get_or_initialize_state` is minimal and adding hooks would be over-engineering. I would consider it if there were more significant buildout for NorMuon variants that would lead to excessive code repetition. 

## Minor
1. Fixed
2. Fixed
3. Preserved to match `NorMuon`. Additionally, the betas for `AdamW` are similarly hardcoded.

JohnLangford:
This is giving good results at the 14b scale.

PR #87, opened 2026-06-06
dion2: restore single-rounding post-ortho for tensor-subclass params

JohnLangford:
Follow-up to #81 (fused Triton post-orthogonalize) and #84 (subclass fallback).

#84 fixed an illegal-memory-access crash by routing wrapper-subclass params (the MXFP8 training weight wrapper, DTensor) away from the raw-`data_ptr()` Triton kernel to the eager `dion2_post_orthogonalize`. But that eager path writes the weight-decayed weight and then accumulates the update in a **second pass** — two roundings on the selected rows/columns — whereas the kernel computes `a*x - b*u` in fp32 and writes **once**. So MXFP8 runs (which set `triton_post_ortho=true`, and whose params are wrapper subclasses) silently lost the kernel's single-rounding numerics.

### Change

`dion2_post_orthogonalize_fused`: gathers the selected slices, computes `(1 - base_lr*wd)*x - adjusted_lr*u` in float32, and writes them once (matching the kernel's single rounding); unselected entries get the weight decay in place. It uses only `__torch_dispatch__`-routed ops (`gather`, `mul_`, `scatter_`) — no raw `data_ptr()` writes — so it is subclass-safe and handles 3D batched params (per-row indices). The #84 wrapper-subclass fallback now routes here; plain dense tensors still take the Triton kernel.

### Why fp32 is not a regression (it's faster)

Single rounding requires a higher-than-storage intermediate — exactly what the Triton kernel already does. Only the `k` selected slices are materialized in fp32. The fused path is **one compiled kernel** vs the eager path's in-place `mul_` plus a per-parameter `scatter_add_` loop, so it is faster despite the fp32:

| post-ortho over full matrix set | eager (two-rounding) | fused (single-rounding) |
|---|---|---|
| 1b | 8.55 ms | 6.18 ms (**−28%**) |
| 4b | 35.8 ms | 26.4 ms (**−26%**) |

(single GPU, plain bf16 proxy; real MXFP8 adds dispatch + requantization overhead to both paths, but the fused path writes X once vs twice, so it should win by more.)

### Numerical benefit

The single-rounding update lowers selected-row update RMSE vs an fp32 reference by ~15% in bf16 (MXFP8 is coarser → larger).

### Tests

`test_post_ortho_triton_falls_back_for_wrapper_subclass` (2D+3D, fp32+bf16): subclass fallback matches `dion2_post_orthogonalize_fused` bit-exactly. `test_fused_eager_matches_kernel_single_rounding` (2D+3D): fused matches the Triton kernel — unselected bit-identical, selected within bf16 rounding — and differs from the two-rounding eager path on the selected slices.

### DIFF
--- a/dion/dion2.py
+++ b/dion/dion2.py
@@ -445,6 +445,49 @@ def dion2_post_orthogonalize(
         x.scatter_add_(dim=select_dim, index=idx_exp, src=u_scaled)
 
 
+@torch.compile(fullgraph=True)
+def dion2_post_orthogonalize_fused(
+    X: List[Tensor],
+    U: List[Tensor],
+    indices: List[Tensor],
+    base_lr: Tensor,
+    adjusted_lr: Tensor,
+    weight_decay: Tensor,
+    select_dim: int,
+):
+    """
+    Single-rounding weight decay + weight update after orthogonalization.
+
+    Computes the new value of the selected rows/columns as
+    ``(1 - base_lr*weight_decay)*x - adjusted_lr*u`` in float32 and writes it
+    once, matching the single-rounding numerics of the fused Triton kernel
+    (dion2_post_orthogonalize_triton). Unselected entries get the weight decay
+    in place, also a single rounding. This differs from dion2_post_orthogonalize,
+    which writes the weight-decayed weight and then accumulates the update in a
+    second pass, rounding the selected slices twice.
+
+    Only the selected slices are gathered into float32, so the extra work over
+    the in-place weight decay is small. Uses only ``__torch_dispatch__``-routed
+    ops (no raw data_ptr writes), so it is safe for traceable wrapper subclasses
+    such as the MXFP8 training weight wrapper, for which the Triton kernel cannot
+    be used. Inputs should be lists of regular Tensor, not DTensor.
+    """
+    a = 1 - base_lr * weight_decay
+    neg_lr = -adjusted_lr
+    for x, u, idx in zip(X, U, indices):
+        if select_dim == -2:
+            idx_exp = idx.unsqueeze(-1).expand_as(u)
+        else:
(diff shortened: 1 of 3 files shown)


JohnLangford:
## Cluster smoke A/B — real MXFP8 (sparse-3b + NorDion2, 32 GPU, `triton_post_ortho=true`)

Two runs differing only in the dion pin: **base** = `main` (#84 two-rounding fallback) vs **fused** = this branch (single-rounding `dion2_post_orthogonalize_fused`). climbmix, ~40 steps after compile.

**Verification:** both ran to completion, no crash; the fused arm exercises the new path on real `MXFP8TrainingWeightWrapperTensor` params. Loss sane and comparable (base 8.03→7.34, fused 7.96→7.23 over the 4 logged steps).

**Timing (median over post-compile steps):**

| metric | base (2-round) | fused (1-round) | Δ |
|---|---|---|---|
| optimizer_time_sec | 0.0704 | 0.0716 | +1.8% |
| step_time_sec | 0.584 | 0.582 | −0.3% |
| tokens/s/GPU | 55.6k | 55.6k | ~0 |
| MFU | 0.507 | 0.507 | ~0 |

**Correction to the description below:** the −28%/−26% figures are the post-ortho *op in isolation* on plain bf16, single GPU. On real MXFP8 that op-level win does **not** carry over — under wrapper-subclass `__torch_dispatch__` + requantization the fused path costs ~1.3 ms more per optimizer step (within-arm spread is <0.5 ms, so this is real, not noise), but it is invisible at the step level (step time / tokens-per-sec / MFU unchanged). The change should be taken on the **single-rounding numerics** (recovers the kernel's accuracy for MXFP8, which #84 silently dropped) at **negligible throughput cost** — not as a speedup.

PR #90, opened 2026-06-12
Add split_sizes param-group option + megabatch cpu overhead optimizations (2x faster cpu time) 

alint77:
## Motivation

A fused QKV projection keeps the model on a single wide GEMM (and plays well with fp8 paths that own one fused weight), but Muon orthogonalizes it as one tall matrix, blending the Q, K, and V projections. Splitting the parameter recovers per-projection Newton-Schulz — I've measured it improving loss — at the cost of narrower GEMMs.

The existing `num_heads` option doesn't cover this case: under grouped-query attention the Q and K/V blocks have **unequal** sizes, and the `num_heads` local-shard path requires each FSDP rank to hold whole blocks, which fails when `kv_dim` exceeds the per-rank shard.

## What this adds

A `split_sizes` param-group option for **Muon** and **NorMuon** that partitions dim 0 of a 2D weight into row blocks and runs Newton-Schulz independently per block:

```python
param_groups = [
    dict(params=fused_qkv_params, split_sizes=(q_dim, kv_dim, kv_dim)),
    dict(params=other_matrix_params),
]
```

Each block receives the same update it would as a separate parameter:

- **Newton-Schulz** runs per block; blocks of equal size (e.g. K and V) are stacked into one batched NS call.
- **LR adjustment** (`spectral_norm` / `rms_norm`) is converted from the whole-matrix value to the per-block value via a post-NS rescale (`adjust(block_shape) / adjust(full_shape)`), so `muon_update_post_orthogonalize` is unchanged.
- **NorMuon's** norm-preserving rescale runs per block on the non-sharded path. The sharded all-to-all path keeps the existing per-shard normalization (block boundaries aren't locally available there), consistent with current distributed behavior.

The split happens on the **fully assembled** matrices after the FSDP all-to-all, so the communication pattern, momentum state layout, and checkpoint format are all unchanged from the fused parameter. Total NS FLOPs decrease relative to fused (smaller matrices).

## Constraints

- 2D parameters only; `split_sizes` must sum to dim 0.
- With FSDP, dim 0 must be divisible by the world size (otherwise shard padding would intersperse zero rows between rank segments of the assembled matrices); raises `NotImplementedError` otherwise.
- Mutually exclusive with `num_heads` and `flatten=True`.
- Dion2 / NorDion2 raise `NotImplementedError` for the option (their orthogonalization operates on local shards).

## Testing

New `TestSplitSizes` class (21 tests): parity against separate per-block parameters for Muon and NorMuon across `adjust_lr` modes and nesterov, megabatching, fused momentum state, and validation errors. Full suite passes on a single GH200 (229 passed, 11 skipped multi-GPU).

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,6 +260,22 @@ When `num_heads > 1`, the optimizer views each 2D weight as a batch of `num_head
 
 Requirements: the parameter must be 2D, `num_heads` must divide dim 0, and when using FSDP it must also divide the world size. Place Q / K / V / gate projections in one group (axis-0 heads); O-projection heads are on axis 1 and are not covered by this option.
 
+### Per-Block Newton-Schulz for Fused QKV Projections
+
+A fused QKV weight of shape `(q_dim + kv_dim + kv_dim, in_features)` keeps the model on a single wide GEMM, but orthogonalizing it as one matrix blends the Q, K, and V projections — and unlike per-head splitting, the blocks may have unequal sizes under grouped-query attention. Muon and NorMuon can run Newton-Schulz independently per row block by setting `split_sizes` on the parameter group:
+
+```python
+param_groups = [
+    dict(params=fused_qkv_params, split_sizes=(q_dim, kv_dim, kv_dim)),
+    dict(params=other_matrix_params),
+    ...
+]
+```
+
+Each block receives the same update it would as a separate parameter: Newton-Schulz, the learning-rate adjustment (`spectral_norm` / `rms_norm`), and NorMuon's norm-preserving rescale are all computed per block. Blocks of equal size (e.g. K and V) are batched into one Newton-Schulz call. The split happens on the fully assembled matrices after the FSDP all-to-all, so the communication pattern is unchanged from the fused parameter.
+
+Requirements: the parameter must be 2D, `split_sizes` must sum to dim 0, and with FSDP dim 0 must be divisible by the world size (so the assembled matrices contain no padding rows). `split_sizes` is mutually exclusive with `num_heads`. With FSDP, NorMuon's norm-preserving rescale operates on local shards (the existing distributed behavior) rather than per block.
+
 ## Distributed Training Configuration
 
 For our efficient distributed optimizers to work correctly, they need information about the model's parallelization scheme. This is provided by passing `DeviceMesh` objects during optimizer construction.
--- a/dion/dion2.py
+++ b/dion/dion2.py
@@ -149,6 +149,10 @@ class Dion2(DistributedOrthoBase):
                 shape_groups[(p.shape, sharding, p.dtype)].append(p)
 
             num_heads = self._resolve_num_heads(group)
+            if group.get("split_sizes") is not None:
+                raise NotImplementedError(
+                    "split_sizes is currently supported only by Muon and NorMuon."
+                )
 
             for (_shape, _sharding, _dtype), params in shape_groups.items():
                 gradients = [p.grad for p in params]
--- a/dion/megabatch_base.py
+++ b/dion/megabatch_base.py
@@ -212,6 +212,66 @@ class DistributedOrthoBase(Optimizer):
             )
         return num_heads
 
+    def _resolve_split_sizes(self, group: dict) -> Optional[Tuple[int, ...]]:
+        """Validate the ``split_sizes`` option on a param group.
+
+        ``split_sizes`` partitions dim 0 of a 2D weight into row blocks (e.g. a
+        fused QKV projection into its Q, K, and V blocks, which may have
+        unequal sizes under GQA). Newton-Schulz and the learning-rate
+        adjustment then run independently per block, matching the update that
+        separate per-block parameters would receive, while the parameter
+        itself stays fused for a single wide GEMM in the model.
+
+        Returns the validated sizes as a tuple, or ``None`` when the option is
+        unset. Raises ``ValueError`` for invalid values or incompatible
+        combinations. The sizes must sum to dim 0 of every parameter in the
+        group; that is checked at task-creation time when shapes are known.
+        """
+        split_sizes = group.get("split_sizes")
+        if split_sizes is None:
+            return None
+        if not isinstance(split_sizes, (tuple, list)) or len(split_sizes) < 2:
+            raise ValueError(
+                f"split_sizes must be a tuple or list of at least 2 block sizes, "
+                f"got {split_sizes!r}."
+            )
+        # bool is a subclass of int in Python; reject it explicitly.
+        if any(
+            isinstance(s, bool) or not isinstance(s, int) or s < 1
+            for s in split_sizes
+        ):
+            raise ValueError(
+                f"split_sizes entries must be positive integers, got {split_sizes!r}."
+            )
+        if self._resolve_num_heads(group) is not None:
+            raise ValueError(
+                "split_sizes is incompatible with num_heads > 1: use num_heads "
+                "for uniform per-head splits or split_sizes for uneven row "
+                "blocks, not both."
+            )
+        if group.get("flatten"):
+            raise ValueError(
+                "split_sizes is incompatible with flatten=True: split_sizes "
+                "applies to 2D parameters only."
+            )
+        return tuple(split_sizes)
+
+    def _validate_split_shape(
+        self, split_sizes: Tuple[int, ...], params: List[Tensor]
+    ) -> None:
+        """Check that ``split_sizes`` is consistent with a shape group's params."""
+        shape = params[0].shape
+        if len(shape) != 2:
+            raise ValueError(
+                f"split_sizes is only supported for 2D parameters, got shape "
+                f"{tuple(shape)}."
+            )
+        if sum(split_sizes) != shape[0]:
+            raise ValueError(
+                f"split_sizes {split_sizes} must sum to dim 0 of the parameter "
+                f"(got shape {tuple(shape)})."
+            )
+
     def _prepare_head_split(
         self,
         num_heads: int,
@@ -396,6 +456,8 @@ def megabatch_orthogonalize_async(
     flatten: bool,
     epsilon: Tensor,
     global_comm_dim_size: Optional[int],
+    split_sizes: Optional[Tuple[int, ...]] = None,
+    split_scales: Optional[Tuple[float, ...]] = None,
 ) -> Generator[None, None, List[Tensor]]:
     """
     Shared megabatch communication + Newton-Schulz orthogonalization.
@@ -422,6 +484,12 @@ def megabatch_orthogonalize_async(
             (``param.shape[comm_dim]``). Used to compute
             ``padded_local_size = ceil(global / world_size)`` so the
             alltoall sees uniform per-pair sizes across ranks.
+        split_sizes: Optional row-block sizes for dim -2. Newton-Schulz runs
+            independently per block (on the fully assembled matrices), as if
+            each block were a separate parameter.
+        split_scales: Optional per-block rescaling applied after Newton-Schulz,
+            used to convert the caller's whole-matrix learning-rate adjustment
+            into the per-block adjustment.
     """
     N = len(U)
 
@@ -460,6 +528,19 @@ def megabatch_orthogonalize_async(
             )
         padded_local_size = (global_comm_dim_size + world_size - 1) // world_size
         original_local_size = U_work[0].size(comm_dim)
+        if (
+            split_sizes is not None
+            and comm_dim == -2
+            and padded_local_size * world_size != global_comm_dim_size
+        ):
+            # Shard padding would intersperse zero rows between rank segments
+            # of the assembled matrices, scrambling the row-block boundaries.
+            raise NotImplementedError(
+                f"split_sizes requires dim 0 of the parameter "
+                f"({global_comm_dim_size}) to be divisible by the sharding "
+                f"world_size ({world_size}) so the assembled matrices have no "
+                f"interspersed padding rows."
+            )
         if padded_local_size < original_local_size:
             raise RuntimeError(
                 f"padded_local_size ({padded_local_size}) < this rank's "
@@ -493,6 +574,8 @@ def megabatch_orthogonalize_async(
             newton_schulz_func=newton_schulz_func,
             flatten=flatten,
             epsilon=epsilon,
+            split_sizes=split_sizes,
+            split_scales=split_scales,
         )
 
         split_chunks = [
@@ -525,6 +608,8 @@ def megabatch_orthogonalize_async(
             newton_schulz_func=newton_schulz_func,
             flatten=flatten,
             epsilon=epsilon,
+            split_sizes=split_sizes,
+            split_scales=split_scales,
         )
 
         all_chunks = [torch.empty_like(my_matrices) for _ in range(world_size)]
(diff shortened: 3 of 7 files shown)


JohnLangford:
This is John's reviewbot.

Reviewed the full diff against the source. This is a clean, well-reasoned patch — the core insight that you can split *after* the all-to-all assembly (keeping comm/momentum/checkpoint layout identical to the fused param, with the only constraint being "no interspersed padding rows") is exactly right, and the divisibility guard enforces it correctly. A few substantive points:

**Correctness — the math checks out, including the subtle parts.**
- Block reassembly under FSDP is sound: contiguous chunking + `dim0 % world_size == 0` ⇒ `padded_local_size == local_size` ⇒ `cat(output_chunks, dim=-2)` reproduces the original row order, so block boundaries land where they should. The `comm_dim == -2` guard at `megabatch_base.py:531` is the right gate.
- The `split_scales = adjust(block)/adjust(full)` trick (pre-divide so the single whole-matrix `adjusted_lr` applied in `post_orthogonalize` lands per-block) is neat and keeps `muon_update_post_orthogonalize` untouched. Verified net LR per block = `base_lr · adjust(block)`, matching a separate param.
- NorMuon parity is **approximate, not exact**, even on a single GPU at fp32 — and not for the bf16 reason the test comment gives. With `V` init=0 and a constant per-block scale `c`, induction gives `V' = c²·V` exactly, but `denom = sqrt(V) + 1e-8` doesn't commute with `c` (`c·sqrt(V)+1e-8 ≠ c·(sqrt(V)+1e-8)`). It's negligible for well-scaled updates (and `rtol=1e-2` absorbs it), but the `normuon.py` comment claiming the scales "commute through the normalization self-consistently" overstates it. Worth a one-line honest caveat about the epsilon term.

**Test coverage — the headline path is untested.** The whole motivation is "num_heads breaks under FSDP, split_sizes doesn't," yet every new test is single-GPU (`CUDA_AVAILABLE`, no `world_size>1`). The divisibility `NotImplementedError`, the assembled-matrix block alignment, and NorMuon's documented per-shard fallback all rest purely on reasoning. The 21 parity tests are good for the local path, but a 2-GPU test (even one in the multi-GPU-skipped tier) exercising the sharded reassembly would meaningfully de-risk the feature that justifies the PR. At minimum, add a unit test asserting the divisibility guard fires (construct the sharded condition directly), since that branch currently has zero coverage.

**Design / possible simplification.** The `split_scales` ratio exists only because `post_orthogonalize` will later multiply by `adjusted_lr(full)`. An alternative that avoids the ratio (and the `/full_adjust` division, and the per-block Python multiply loop): build a single `[rows,1]` per-row LR vector = `adjust(block_i)` broadcast over each block, and apply it as one vectorized multiply on the assembled matrix — no dependence on the full-matrix adjustment at all. It's arguably cleaner and composes with NorMuon identically, at the cost of touching the post-ortho LR application. The current minimal-touch approach is a defensible trade; just noting it. Threading `split_sizes`/`split_scales` through 5 call sites is verbose but localized and all-optional.

**Backward compat / security:** clean. All new kwargs default to `None`; the new param-group key is opt-in; Dion2/NorDion2 reject it explicitly. No security surface (pure numerics, no external input). Bool-rejection in validation is a nice touch.

**Perf/throughput:** FLOPs-down claim is correct (`Σ rᵢ² ≤ (Σ rᵢ)²`), and batching equal-height blocks (K+V) into one NS call is valid since NS is per-batch-element independent. Net wall-clock may not improve on small matrices (more kernel launches + `stack`/`cat`/`split` allocs each step), but the patch sells this on loss quality, not speed, so that's fine.

Nits (non-blocking): the sum-mismatch validation fires at first `step()` rather than construction — consistent with `num_heads`, fine; a group carrying `split_sizes` with `algorithm=adamw/lion` silently ignores it (a warning would be friendlier).

Solid work overall — main asks are (1) the epsilon caveat in the NorMuon comment and (2) some automated coverage of the sharded path or at least the divisibility guard.

alint77:
This feature is mostly aimed at fp8/fp4 training. the bigger qkv matmul amortizes the quantization costs much better compared to separate q,k,v matmuls which would have their own separate quantization/amax kernels that might even cause slowdown with fp8.

alint77:
and the same can be applied to gated MLPs, where Wi is separated into 2 matrices (up, gate), and doing separate ortho on those also improves loss.

JohnLangford:
Do you have performance impact summary (computational & statistical)?

NoahAmsel (commented):
I don't know whether the runtime benefits will be significant, but many people have existing codebases with merged QKV and this could make dion easier to adopt. (on the other hand, they are already covered by split heads unless they use grouped query attention). still, i agree it would be great to get a sense of whether this is beneficial in practice (even just for aifsdk)

my only concern is for the added complexity. it is awkward to have even one system for reshaping weight matrices, let alone two totally separate systems. in terms of the implementation: in this PR, _newton_schulz_row_blocks is responsible for doing the slicing / reshaping, but for num_heads, this reshaping happens elsewhere. isn't it better to have it all in one place?

likewise is there anyway to simplify the interface? perhaps we could have a single parameter where you either give the sizes of the blocks explicitly (as in this PR) or you give a number of blocks and it divides the matrix into equal sized pieces (as with num_heads) or perhaps at least some of the rescaling logic can be shared?

alint77:
split head doesn't work when:
1- world_size%3 != 0 in fsdp mode
2- gqa, pairedheadattn or diffv2 where q,k,v have different sizes

alint77:
h1024 16layer 170M model, FP8 on Wqkv and MLP, 4xH100 (4xGH200) fsdp2:

Fused Wqkv: 179ms
Unfused Wqkv: 186ms

same model with Hdim 2048:

424ms
451ms

alint77:
breakdown of kernels:


run | NorMuon kernel launches | summed kernel time | NCCL launches | NVJit GEMM launches | Triton launches
-- | -- | -- | -- | -- | --
split | 1064 | 33.03 ms | 32 | 384 | 440
whole | 788 | 29.38 ms | 24 | 288 | 332

New split mode is naturally a bit slower when looking at the opt.step alone, but this is offset by the gains in fused Wqkv matmul as shown in the e2e training numbers in the post above

alint77:
<p>At the logged matched steps, <strong>split ortho is consistently a little better than whole-<code>Wqkv</code> ortho</strong>.</p><p><strong>Eval Loss</strong></p><div><div>

comparison | step | split | whole | whole - split
-- | -- | -- | -- | --
4-node | 1000 | 2.5908 | 2.5919 | +0.0011
4-node | 2000 | 2.5226 | 2.5261 | +0.0035
4-node | 3000 | 2.4818 | 2.4858 | +0.0040
4-node | 4000 | 2.4501 | 2.4552 | +0.0051
4-node | 8000 | 2.3863 | 2.3917 | +0.0054

keep in mind this was just a quick AB test at only 8B tokens (global_BS=1M), but the improvement is consistant on every combination of features/sizes that I've tested. 

In short, the point of this feature is to get the best of both worlds instead of having to choose one: faster gemms due to fused qkv + better loss due to split NS

alint77:
Pushed a follow-up that trims host-side (CPU) overhead in the NorMuon step.

**Motivation.** At scale the NorMuon `opt.step` is host-bound: the GPU finishes its ~21 ms of kernels partway through the step and then sits idle while the rest of the CPU dispatch plays out. A lot of that dispatch isn't optimizer math — it's the megabatch plumbing issuing one small op per parameter/shard. So reducing the host op count directly shortens the step whenever the command queue drains.

**Changes (both bit-identical, no numerics change):**
- `normuon.py` — write the variance buffers back with a single `torch._foreach_copy_` and `unbind` U in one call, instead of N per-tensor `copy_`/select launches.
- `megabatch_base.py` — vectorize the all-to-all pack/unpack in `megabatch_orthogonalize_async`. The per-`(rank × per_rank)` Python list comprehensions of `stack`/`select`/`narrow`/`contiguous` become a single `stack` + view + `unbind`, so the expensive memory ops fire once per shape-group rather than once per parameter.

**Results** (33-layer model, 164 muon params, world=16, measured on a queue-drained `opt.step`):

| metric | before | after | |
|---|---:|---:|---|
| NorMuon `opt.step` CPU time | 86.6 ms | 41.0 ms | **2.1×** |
| CPU ops / step | 3,688 | 2,931 | −21% |
| GPU time / step | 21.5 ms | 21.3 ms | unchanged |

Where it comes from — the expensive per-parameter ops collapse to per-shape-group, while the cheap metadata is unchanged:

| op | before | after |
|---|---:|---:|
| `aten::copy_` | 251 | 54 |
| `aten::narrow` | 193 | 5 |
| `aten::slice` | 257 | 69 |
| `aten::cat` | 85 | 29 |
| `aten::stack` | 80 | 24 |

Numerics are unchanged (loss matches to 4 decimals across a run); existing optimizer + empty-shard tests pass, and I validated the sharded all-to-all path end-to-end across 4 gloo ranks (including the padding-only / empty-shard cases).

@JohnLangford @NoahAmsel feedbacks appreciated

JohnLangford:
This is John's reviewbot.

Re-reviewed for the new `93fad5a` ("Reduce NorMuon host overhead"). I traced all five edits against the old code; the split_sizes commit I covered in my earlier comment, so this focuses on the host-overhead work.

**Correctness — the "bit-identical" claim holds up.** Every edit is a pure data-movement reshuffle with the arithmetic kernels untouched, and the ordering is preserved in each case:
- `input_chunks`: `stack(U_work).unflatten(0,(world_size,per_rank)).unbind(0)` reproduces the old per-rank `stack(U_work[r*per_rank:(r+1)*per_rank])` exactly (row-major ⇒ chunk `[r][i] = U_work[r*per_rank+i]`), and first-dim unbind of a freshly-stacked contiguous tensor yields contiguous views — safe to hand to `all_to_all`.
- recv unpack: `stack(recv_chunks).narrow(comm_dim,…).contiguous().flatten(0,1).unbind(0)` — `flatten(0,1)` keeps the rank-major/per_rank-minor order of the old double loop, and `comm_dim` being negative still indexes the matrix dim after the two leading axes are added. Empty/padding-only ranks (`original_local_size==0`) degrade to empty slices as before.
- `_foreach_copy_(V_local, V_stacked.unbind(0))` + `U_stacked.unbind(0)` are exact equivalents of the per-i `copy_`/select loop.

So bit-identical is credible — the "matches to 4 decimals" framing actually understates it; these ops don't reorder any float math.

**Two substantive notes:**

1. **The all-gather path now adds a copy it didn't have.** In the `N>1, process_group, comm_dim is None` branch, the old `[all_chunks[r][i] …]` returned zero-copy views into the gather buffers; the new `stack(all_chunks).flatten(0,1).unbind(0)` materializes a full copy of the gathered megabatch. It's a defensible trade (fewer host dispatches for one device memcpy, consistent with the host-bound goal), but unlike the sharded recv path — where the old code already did a per-element `.contiguous()` — here you're introducing a copy that wasn't there. Worth confirming the dispatch saving outweighs it on this specific (replicated/batch-sharded) path, or leaving it as views.

2. **`torch._foreach_copy_` is a private API.** Fine given dion already leans on `torch._foreach_*`, just noting the surface.

**Test coverage — the real gap, and it's now closeable in-CI.** Three of the four `megabatch_base` edits (input pack, recv unpack, all-gather unpack) live on paths that only execute with `world_size>1`, so single-GPU CI exercises none of them — the rewritten sharded pack/unpack ordering, the highest-risk part of this commit (and the same path the split_sizes divisibility guard depends on), has zero committed regression coverage. You say you validated 4 gloo ranks end-to-end — that's exactly the right test; please **commit it**. `tests/test_megabatch_empty_shard.py` already has an `mp.spawn` harness, and since gloo runs on CPU it would execute in normal CI without needing GPUs. A small gloo-backed `mp.spawn` parity test (assemble → split → reassemble equals the reference, incl. the padding-only case) would lock in the ordering invariant permanently and catch any future regression in this plumbing.

**Perf / backward compat:** 2.1× host-time on `opt.step` (86.6→41 ms, GPU flat) is a clean, believable win from collapsing per-param dispatch to per-shape-group, and the megabatch_base changes benefit Muon too, not just NorMuon. Bit-identical ⇒ no compat or checkpoint concerns. Nice optimization — main ask is to land the gloo validation as a committed CI test rather than a one-off.

alint77:
**Follow-up: skip the unbind→restack round-trip.** NorMuon re-stacks the orthogonalized update for its normalization right after `megabatch_orthogonalize_async` hands back a list — so the megabatch's `unbind` and the caller's `torch.stack` cancel out into a pure round-trip (N selects + a stacking copy per shape group). I added an opt-in `return_stacked` flag to `megabatch_orthogonalize_async` that returns the already-assembled `[N, *shape]` tensor instead of a list, and NorMuon now consumes it directly. That trims another **~13–17%** of the step's CPU ops (`select`/`stack`/`unbind`/`as_strided`). Muon and Dion2 keep the list contract (`return_stacked=False` default, unchanged); **NorDion2 has the exact same restack pattern, so it can opt in the same way** — I left it as-is here to keep the change small, but it's a one-line follow-up. Same bit-identical / gloo-validation as above. (Worth noting these removed ops are cheap metadata, so this one is an op-count cleanup so the gains in terms of wallclock is pretty minimal)

JohnLangford:
This is John's reviewbot.

Reviewed the new `0333cc1` ("Skip unbind->restack round-trip"). My earlier comments cover split_sizes and the host-overhead commit; this is a small, clean follow-up and it checks out.

**Correctness — bit-identical holds across all four branches.** I traced each:
- `_finalize(flat)` does `flat[:N]` then stacked-or-`unbind`; with `return_stacked=False` that's exactly the old `list(...unbind(0))[:N]` (slicing first-N commutes with unbind), so Muon/Dion2/single-GPU paths are untouched.
- `return_stacked=True` now hands NorMuon a contiguous `[N,*shape]` **view** of the comm buffer instead of the old `torch.stack(U)` copy. I checked `normuon_normalization_stacked` / `_normuon_normalization_core` never mutate `U` in place — every op (`U*U`, `U/denom`, the Frobenius rescale) allocates, and the normalized output is always a fresh tensor — so the aliasing into `recv_stacked` is safe and the buffer stays alive via the view's refcount. No use-after-free, no corruption.
- N==1 (`unsqueeze(0)`) and the `else`/all-gather paths are value-identical to the old list→`torch.stack`, minus the copy. Nice side effect: this removes the extra stack-copy I flagged on the all-gather path last round — NorMuon now consumes that assembled tensor directly.

**Design.** The opt-in flag is the right minimal choice — Muon/Dion2 keep the list contract, only the caller that immediately re-stacks opts in. `_finalize` also de-duplicates the four return sites cleanly. No more elegant alternative worth the churn (always-stacked would force unbinds into Muon/Dion2 for no gain).

**Coverage.** The NorMuon `return_stacked=True` path on the *non-sharded* branches (`else`, N==1) is exercised by existing single-GPU NorMuon/split_sizes tests, so this commit's hot path has real CI coverage. The *sharded* return_stacked variant is still gloo-manual-only — which reinforces last round's ask: please land your `stacked==list` 4-rank gloo check as a committed `mp.spawn`/gloo test in `tests/test_megabatch_empty_shard.py` (gloo runs on CPU, so it gates in CI). That one test would now lock in three things at once: the sharded pack/unpack ordering, the split_sizes divisibility path, and this return_stacked equivalence.

**Perf / compat.** ~13-17% fewer host ops with GEMM/copy work unchanged is a believable, free win; bit-identical ⇒ no checkpoint/compat impact. Good incremental cleanup.

PR #93, opened 2026-06-22
Make gram-newton-schulz / quack-kernels optional extras

gugarosa:
## What

Move `gram-newton-schulz` and `quack-kernels` from the mandatory `requirements_dion.txt` into an optional `dion[gram-newton-schulz]` extra (aliased `gns`). The `dev` and `train` extras stay GNS-free too, so the default `pip install -e .[train]` is FA4/Blackwell-clean; only `[gns]` (or `[train,gns]` / `[dev,gns]`) pulls the kernels.

## Why

These packages are only used by the opt-in Gram Newton-Schulz orthogonalization path (`use_gram_newton_schulz=True`, which defaults to `False`). The import is already lazy and guarded in `dion/megabatch_base.py`:

```python
elif use_gram_newton_schulz:
    try:
        from gram_newton_schulz import GramNewtonSchulz
    except ImportError:
        raise ImportError(
            "use_gram_newton_schulz=True requires the 'gram-newton-schulz' package, "
            "which is not installed. "
            "Install it with: pip install gram-newton-schulz"
        )
```

But as **hard** dependencies they are installed for everyone, and `gram-newton-schulz` transitively hard-pins `nvidia-cutlass-dsl==4.4.2`. On modern Flash-Attention-4 / Blackwell (B200) images that ship `nvidia-cutlass-dsl==4.5.2`, a plain `pip install dion` (or `pip install -e .` of a project that depends on dion) silently downgrades cutlass to `4.4.2` and breaks FA4. Consumers currently work around this with `--no-deps`, which is fragile and easy to get wrong.

Making the kernels an extra keeps the base install agnostic to the host CUDA-kernel stack. Users who want GNS opt in with the extras-from-git form (since `dion` is not on PyPI):

```bash
pip install "dion[gram-newton-schulz] @ git+https://github.com/microsoft/dion.git"
# or, from a clone:
pip install -e ".[gram-newton-schulz]"
```

Note: the extra still pins `nvidia-cutlass-dsl==4.4.2`, so opting in on a Flash-Attention-4 / Blackwell image (cutlass `4.5.2`) re-introduces that conflict — use a separate environment if you need both.

## Notes

- The `dev` and `train` extras stay GNS-free so the default `pip install -e .[train]` is FA4/Blackwell-clean; the GPU-gated `use_gram_newton_schulz=True` tests `pytest.importorskip("gram_newton_schulz")` and skip when the extra is absent. Use `pip install -e ".[train,gns]"` to train with the kernels.
- `quack-kernels` has no direct imports in `dion/` (it is only pulled in transitively by `gram-newton-schulz`), so both move together.
- The base optimizers (`Dion`, `Dion2`, `Muon`, `NorMuon`) import only torch/stdlib at module load, so `from dion import ...` is unaffected by this change.
- The only code change is a one-line `ImportError` message in `megabatch_base.py` pointing at the optional package / extra; the lazy import already covers the "opted in but not installed" case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,21 @@
+# Changelog
+
+All notable changes to this project are documented in this file.
+
+## [Unreleased]
+
+### Changed
+
+- **Breaking (install):** `gram-newton-schulz` and `quack-kernels` are no longer
+  base dependencies. They moved to an optional `dion[gram-newton-schulz]` extra
+  (alias `dion[gns]`), and are also excluded from the `dev` and `train` extras.
+  This keeps the default install free of the transitive `nvidia-cutlass-dsl==4.4.2`
+  pin, which conflicts with Flash-Attention-4 / Blackwell stacks built on cutlass
+  `4.5.2`.
+
+  **Action required:** if you run with `use_gram_newton_schulz=True`, install the
+  extra (`pip install "dion[gns] @ git+https://github.com/microsoft/dion.git"`, or
+  `pip install -e ".[gns]"` from a clone). Without it, optimizer construction now
+  raises a clear `ImportError` at runtime instead of the kernels being silently
+  present. Opting in re-introduces the cutlass `4.4.2` pin, so use a separate
+  environment from FA4/Blackwell.
--- a/README.md
+++ b/README.md
@@ -68,6 +68,7 @@ git clone https://github.com/microsoft/dion.git
 cd dion
 pip install -e .[train]
 ```
+> `train` stays free of the Gram Newton-Schulz kernels (and their `nvidia-cutlass-dsl==4.4.2` pin) so the default training install works on Flash-Attention-4 / Blackwell stacks. To train with `--use_gram_newton_schulz`, use `pip install -e ".[train,gns]"` in a separate environment.
 
 Download pretokenized FineWeb dataset:
 ```bash
--- a/README.md
+++ b/README.md
@@ -50,6 +50,8 @@ Our implementations are available as a `pip` package! Install to use in your pro
 pip install git+https://github.com/microsoft/dion.git
 ```
 
+> The optional Gram Newton-Schulz orthogonalization kernels (enabled with `use_gram_newton_schulz=True`) are not pulled in by the base install. Add them with `pip install "dion[gram-newton-schulz]"`.
+
 Then in your code, you can use:
 
 ```python
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Our implementations are available as a `pip` package! Install to use in your pro
 pip install git+https://github.com/microsoft/dion.git
 ```
 
-> The optional Gram Newton-Schulz orthogonalization kernels (enabled with `use_gram_newton_schulz=True`) are not pulled in by the base install. Add them with `pip install "dion[gram-newton-schulz]"`.
+> The optional Gram Newton-Schulz orthogonalization kernels (enabled with `use_gram_newton_schulz=True`) are not pulled in by the base install. Add them with `pip install "dion[gram-newton-schulz] @ git+https://github.com/microsoft/dion.git"`, or `pip install -e ".[gram-newton-schulz]"` from a clone. Note: this extra pins `nvidia-cutlass-dsl==4.4.2`, which conflicts with Flash-Attention-4 / Blackwell stacks built on cutlass `4.5.2`, so install it in a separate environment if you need both.
 
 Then in your code, you can use:
 
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ git clone https://github.com/microsoft/dion.git
 cd dion
 pip install -e .[train]
 ```
-> `train` stays free of the Gram Newton-Schulz kernels (and their `nvidia-cutlass-dsl==4.4.2` pin) so the default training install works on Flash-Attention-4 / Blackwell stacks. To train with `--use_gram_newton_schulz`, use `pip install -e ".[train,gns]"` in a separate environment.
+> `train` stays free of the Gram Newton-Schulz kernels (and their `nvidia-cutlass-dsl==4.4.2` pin) so the default training install works on Flash-Attention-4 / Blackwell stacks. To train with `--use_gram_newton_schulz`, use `pip install -e ".[train,gns]"` in a separate environment. Likewise, to develop or test the Gram Newton-Schulz path, install `pip install -e ".[dev,gns]"` — a plain `[dev]` install skips the GNS-specific test cases.
 
 Download pretokenized FineWeb dataset:
 ```bash
--- a/dion/megabatch_base.py
+++ b/dion/megabatch_base.py
@@ -79,9 +79,10 @@ class DistributedOrthoBase(Optimizer):
                 from gram_newton_schulz import GramNewtonSchulz
             except ImportError:
                 raise ImportError(
-                    "use_gram_newton_schulz=True requires the 'gram-newton-schulz' package, "
-                    "which is not installed. "
-                    "Install it with: pip install gram-newton-schulz"
+                    "use_gram_newton_schulz=True requires the optional 'gram-newton-schulz' "
+                    "package, which is not installed. Install it with: "
+                    'pip install gram-newton-schulz (or reinstall dion with its '
+                    '"gram-newton-schulz" extra).'
                 )
             use_polar_express = True
             _gns = GramNewtonSchulz(
--- a/dion/megabatch_base.py
+++ b/dion/megabatch_base.py
@@ -80,9 +80,8 @@ class DistributedOrthoBase(Optimizer):
             except ImportError:
                 raise ImportError(
                     "use_gram_newton_schulz=True requires the optional 'gram-newton-schulz' "
-                    "package, which is not installed. Install it with: "
-                    'pip install gram-newton-schulz (or reinstall dion with its '
-                    '"gram-newton-schulz" extra).'
+                    'package, which is not installed. Install it with: pip install -e ".[gns]" '
+                    "(or pip install gram-newton-schulz)."
                 )
             use_polar_express = True
             _gns = GramNewtonSchulz(
--- a/requirements_dion.txt
+++ b/requirements_dion.txt
@@ -1,4 +1,2 @@
 numpy
 torch>=2.7.1
-gram-newton-schulz==0.1.4
-quack-kernels==0.4.1
--- a/requirements_gns.txt
+++ b/requirements_gns.txt
@@ -1,2 +1,6 @@
+# Optional Gram Newton-Schulz orthogonalization kernels (use_gram_newton_schulz=True).
+# quack-kernels is pulled in transitively by gram-newton-schulz; the explicit pin here
+# is for reproducibility and must track whatever quack version the pinned gram-newton-schulz
+# requires (gram-newton-schulz==0.1.4 also transitively pins nvidia-cutlass-dsl==4.4.2).
 gram-newton-schulz==0.1.4
 quack-kernels==0.4.1
--- a/requirements_gns.txt
+++ b/requirements_gns.txt
@@ -1,0 +1,2 @@
+gram-newton-schulz==0.1.4
+quack-kernels==0.4.1
--- a/setup.py
+++ b/setup.py
@@ -22,14 +22,16 @@ def read_requirements(path):
 # requirements_dion contains the dependencies for the standalone optimizer
 install_requires = read_requirements(req_path)
 
+# requirements_gns contains the optional Gram Newton-Schulz orthogonalization kernels
+gns_requires = read_requirements(req_gns_path)
+
 # requirements_dev contains the dependencies for development, e.g., testing, linting, etc.
-install_dev_requires = install_requires + read_requirements(req_dev_path)
+# GNS is folded in so `dion[dev]` can still exercise the use_gram_newton_schulz tests
+install_dev_requires = install_requires + read_requirements(req_dev_path) + gns_requires
 
 # requirements_train contains the dependencies for training, e.g., datasets, etc.
-install_train_requires = install_requires + read_requirements(req_train_path)
-
-# requirements_gns contains the optional Gram Newton-Schulz orthogonalization kernels
-install_gns_requires = install_requires + read_requirements(req_gns_path)
+# GNS is folded in so train.py's --use_gram_newton_schulz path stays available
+install_train_requires = install_requires + read_requirements(req_train_path) + gns_requires
 
 readme_path = os.path.join(this_directory, "README.md")
 readme_contents = ""
@@ -73,6 +75,8 @@ setup(
         "dev": install_dev_requires,  # Can be installed with `pip install dion[dev]`
         "train": install_train_requires,
         "triton": install_requires + ["triton"],
-        "gram-newton-schulz": install_gns_requires,  # Can be installed with `pip install dion[gram-newton-schulz]`
+        # `pip install "dion[gram-newton-schulz] @ git+https://github.com/microsoft/dion.git"` (alias: gns)
+        "gram-newton-schulz": install_requires + gns_requires,
+        "gns": install_requires + gns_requires,
     },
 )
--- a/setup.py
+++ b/setup.py
@@ -26,12 +26,10 @@ install_requires = read_requirements(req_path)
 gns_requires = read_requirements(req_gns_path)
 
 # requirements_dev contains the dependencies for development, e.g., testing, linting, etc.
-# GNS is folded in so `dion[dev]` can still exercise the use_gram_newton_schulz tests
-install_dev_requires = install_requires + read_requirements(req_dev_path) + gns_requires
+install_dev_requires = install_requires + read_requirements(req_dev_path)
 
 # requirements_train contains the dependencies for training, e.g., datasets, etc.
-# GNS is folded in so train.py's --use_gram_newton_schulz path stays available
-install_train_requires = install_requires + read_requirements(req_train_path) + gns_requires
+install_train_requires = install_requires + read_requirements(req_train_path)
 
 readme_path = os.path.join(this_directory, "README.md")
 readme_contents = ""
--- a/setup.py
+++ b/setup.py
@@ -6,6 +6,7 @@ this_directory = os.path.dirname(__file__)
 req_path = os.path.join(this_directory, "requirements_dion.txt")
 req_dev_path = os.path.join(this_directory, "requirements_dev.txt")
 req_train_path = os.path.join(this_directory, "requirements_train.txt")
+req_gns_path = os.path.join(this_directory, "requirements_gns.txt")
 
 
 def read_requirements(path):
@@ -26,6 +27,9 @@ install_dev_requires = install_requires + read_requirements(req_dev_path)
 
 # requirements_train contains the dependencies for training, e.g., datasets, etc.
 install_train_requires = install_requires + read_requirements(req_train_path)
+
+# requirements_gns contains the optional Gram Newton-Schulz orthogonalization kernels
+install_gns_requires = install_requires + read_requirements(req_gns_path)
 
 readme_path = os.path.join(this_directory, "README.md")
 readme_contents = ""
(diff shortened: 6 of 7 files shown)


JohnLangford:
This is John's reviewbot.

Solid, well-motivated change — decoupling the base install from the `gram-newton-schulz` → `nvidia-cutlass-dsl==4.4.2` transitive pin is the right call, and it's a real win for FA4/B200 images that ship cutlass `4.5.2`. The lazy guard in `megabatch_base.py` already covers the opt-in-but-not-installed case, so the runtime story is clean. A few substantive points before merge:

**1. (Blocker — correctness) The documented install command can't work: `dion` isn't on PyPI.**
The README note and the PR description both tell users to run `pip install "dion[gram-newton-schulz]"`, but `dion` is git-only (the base instruction right above it is `pip install git+https://github.com/microsoft/dion.git`, and `https://pypi.org/pypi/dion/json` 404s). A bare `pip install "dion[...]"` will fail with *"No matching distribution found for dion"* (or, worse, resolve some unrelated PyPI squatter). The extras-from-git form is needed:
```bash
pip install "dion[gram-newton-schulz] @ git+https://github.com/microsoft/dion.git"
# or, from a local clone:
pip install -e ".[gram-newton-schulz]"
```
Please fix both the README note and the PR description.

**2. (Test/dev regression) The `dev` and `train` extras silently lose GNS.**
Both are built as `install_requires + ...`, so removing GNS from `requirements_dion.txt` also drops it from `dion[dev]` and `dion[train]`. After this change, on a CUDA+Triton box `pip install -e .[dev] && pytest` will hit the lazy `ImportError` for the `test_triton_vs_default[...use_gram_newton_schulz=True]` cases (`tests/test_dion2_post_ortho_triton.py:337`) — they're GPU-gated, not GNS-gated, so they'll error rather than skip. Either fold GNS into the dev extra (`install_dev_requires + read_requirements(req_gns_path)`) or guard the test with `pytest.importorskip(\"gram_newton_schulz\")`. Same consideration for `train` since `train.py` exposes `--use_gram_newton_schulz`, though that path is at least guarded at runtime.

**3. (Consistency, cheap) Align the runtime `ImportError` with the new extra.**
`megabatch_base.py:84` still suggests `pip install gram-newton-schulz`. Now that there's a canonical extra, pointing it at `pip install \"dion[gram-newton-schulz]\"` would match the docs. A one-line message tweak — worth it for the UX even though the PR is otherwise code-free.

**4. (Doc caveat) Note that opting in re-introduces the cutlass pin.**
The whole motivation is avoiding the `cutlass-dsl==4.4.2` downgrade, but `dion[gram-newton-schulz]` still drags it in transitively. FA4 users who opt in will re-break FA4 in that env — a one-line README caveat (use a separate env / pin cutlass after) would close the loop.

**Design / elegance (optional, non-blocking):** The new `requirements_gns.txt` + `install_requires +` prefix is heavier than the sibling `triton` extra, which just inlines `install_requires + [\"triton\"]`. You could match that with `\"gram-newton-schulz\": install_requires + [\"gram-newton-schulz==0.1.4\", \"quack-kernels==0.4.1\"]` and drop the new file. It's a wash — the file keeps the pins discoverable and matches dev/train — so either is fine. (Aside: the `install_requires +` in every extra is redundant since pip always installs the base alongside an extra; pre-existing, not introduced here.) A short `gns` alias alongside the long name would also be friendly.

**Backward compat:** The one real break is that existing users who did plain `pip install dion` + `use_gram_newton_schulz=True` now need the extra on a fresh install. That's the intended trade-off and the lazy guard gives a clear error — just worth a release-note / CHANGELOG line since it's technically breaking for that opt-in group.

No performance, throughput, or security concerns — this is packaging metadata only, and it strictly improves install correctness on modern kernel stacks. Fix #1 (and ideally #2) and this is good to go.

gugarosa:
Thanks for the thorough review — all four points addressed in 01287d7.

**1. (Blocker) Install command can't work — `dion` isn't on PyPI.** Fixed in both places. The README note and PR description now use the extras-from-git form:
```bash
pip install "dion[gram-newton-schulz] @ git+https://github.com/microsoft/dion.git"
# or, from a clone:
pip install -e ".[gram-newton-schulz]"
```

**2. (dev/train regression) Extras silently lose GNS.** Folded GNS into both: `install_dev_requires` and `install_train_requires` now append `read_requirements(req_gns_path)`. So `dion[dev]` (and the `test_dion2_post_ortho_triton.py` GNS cases) and `train.py --use_gram_newton_schulz` keep working — only the bare `pip install dion` drops the kernels, which is the goal. (Went with folding-in over `importorskip` so the dev env stays fully capable.)

**3. (Error alignment) `megabatch_base.py:84`.** Updated the `ImportError` to call out the optional package and the extra. Kept the direct `pip install gram-newton-schulz` as the primary hint since that one *is* on PyPI and needs no git URL, and added "or reinstall dion with its `gram-newton-schulz` extra".

**4. (Doc caveat) Opting in re-introduces the cutlass pin.** Added to the README note: the extra still pins `nvidia-cutlass-dsl==4.4.2`, so FA4/Blackwell users who opt in should use a separate environment.

**Design/alias:** Added a short `gns` alias alongside `gram-newton-schulz`. Kept `requirements_gns.txt` since it's now referenced by the extra *and* folded into dev/train (keeps the pins in one place).

**Backward compat:** Agreed it's technically breaking for the `pip install dion` + `use_gram_newton_schulz=True` group. There's no `CHANGELOG` in the repo, so I've called it out in the PR description rather than adding a changelog convention here — happy to add one if you'd prefer.

JohnLangford:
This is John's reviewbot — follow-up on `01287d7f4` ("Address review…").

Thanks for the fast turnaround. All five points from the prior pass are resolved:
- **Install docs** now use the extras-from-git form (`dion[gram-newton-schulz] @ git+…`) and `-e ".[gram-newton-schulz]"` — correct for a non-PyPI package.
- **GNS folded back into `dev`/`train`** (`+ gns_requires`) restores the GPU-gated `use_gram_newton_schulz=True` test coverage and keeps `train.py`'s flag usable.
- **`ImportError`** now points at the extra, and the **`gns` alias** + **cutlass `4.4.2` caveat** in the README are nice touches. The `gns_requires` refactor (read once, reuse) is clean.

One remaining design consideration (non-blocking):

**Folding GNS into `train` reintroduces the cutlass `4.4.2` pin for training installs.** `pip install -e .[train]` now transitively drags in `nvidia-cutlass-dsl==4.4.2` again — i.e. exactly the FA4/Blackwell breakage this PR set out to avoid, for the training workflow. It matches pre-PR behavior so it's not a regression, and the win for base/library consumers stands. But given the PR's own motivation is FA4 training images, consider keeping `train` GNS-free and documenting `[train,gns]` for those who actually want it — that keeps the default training install FA4-clean. Judgment call; fine either way as long as it's intentional.

**More elegant alternative for the `dev` extra (optional):** rather than pulling GPU-only kernels into every `dion[dev]` install just to run two GPU-gated parametrize cases, guard the test with `pytest.importorskip("gram_newton_schulz")` (it's already `skipif(not TRITON_AND_CUDA)`, so this composes naturally) and leave `dev` lightweight. That keeps `dion[dev]` CPU-friendly for lint/format-only contributors. The fold-in you chose is also valid and needs no test changes — just flagging the lighter option.

Neither is a blocker. LGTM to merge as-is; the `train`-extra question is the only thing worth a moment's thought.

gugarosa:
Thanks — took both non-blocking suggestions in 2ad25d7, since they point the same way:

**`train` extra reintroducing the cutlass pin.** Agreed this was self-defeating for the PR's own FA4-training motivation. Reverted both `dev` and `train` to GNS-free (`install_requires + read_requirements(...)`, no `+ gns_requires`), so the default `pip install -e .[train]` stays FA4/Blackwell-clean. Documented `pip install -e ".[train,gns]"` in the README for training installs that actually want the kernels.

**`dev` extra / test coverage.** Went with your `importorskip` suggestion: the GPU-gated `test_triton_vs_default[...use_gram_newton_schulz=True]` cases now `pytest.importorskip("gram_newton_schulz")`, so they skip cleanly when the extra is absent and `dion[dev]` stays CPU-friendly for lint/format-only contributors. The `use_gram_newton_schulz=False` cases are unaffected.

Net: base / dev / train are all kernel-free now; `[gns]` (and its `gram-newton-schulz` alias) is the only thing that pulls `nvidia-cutlass-dsl==4.4.2`. `gns_requires` is still read once and reused by those extras.

JohnLangford:
This is John's reviewbot.

**Verdict:** Solid, well-scoped fix. I verified the core claims hold: GNS is imported lazily only at `megabatch_base.py:79` (so `from dion import Dion2/Muon/NorMuon/...` stays torch/stdlib-only), `quack-kernels` has no direct imports under `dion/`, `test_dion2_post_ortho_triton.py` is the only test touching the GNS path, and `importorskip(..., reason=...)` is valid back to at least pytest 7.4. Base install correctly sheds the `nvidia-cutlass-dsl==4.4.2` pin. 

A few substantive points:

**1. The PR description contradicts the code.** The "What/Why" says *"The `dev` and `train` extras keep them, so only the bare base install drops them."* But the final patch (2ad25d7) makes `dev`/`train` GNS-**free**: they read `requirements_dev.txt`/`requirements_train.txt` (neither has gns) and `setup.py` doesn't add `gns_requires` to them. The code matches the **Notes** section, not the top description. Please fix the stale description so it doesn't mislead release notes.

**2. Test-coverage regression (most important).** With `dev`/`train` GNS-free, the recommended `pip install -e .[dev]` no longer exercises the GNS path — `test_triton_vs_default[...use_gram_newton_schulz=True]` now silently skips. There's no CI workflow in the repo to catch this, so the True-branch becomes effectively untested in the default dev env. Suggest standardizing `pip install -e ".[dev,gns]"` for development and, when CI is added, running one job with the extra so the GNS branch stays covered.

**3. Behavior change for existing users.** Anyone on a base or `[train]` install with `use_gram_newton_schulz=True` will now hit a runtime `ImportError` at optimizer construction instead of it "just working." Well-documented and fine for pre-1.0, but worth an explicit release-note line since the failure moves from install-time to runtime.

**4. Mitigation, not resolution.** The `gns` extra still transitively pins `nvidia-cutlass-dsl==4.4.2`, so GNS and FA4/Blackwell stay mutually exclusive (you note this). The only complete fix is upstream in `gram-newton-schulz` loosening that pin to a 4.5.x-compatible range. Worth tracking as a follow-up; this PR cleanly unblocks the *default* install meanwhile.

**Minor / optional:**
- The new `ImportError` message points at `pip install gram-newton-schulz`, while the README steers users to the extras form — consider aligning (e.g. `pip install -e ".[gns]"`) for one consistent story.
- Two aliases (`gram-newton-schulz` + `gns`) for the same dep set is fine, just doubles the surface.
- `extras_require` entries prepend `install_requires +`, which is redundant (pip always installs base deps with any extra). Harmless and pre-existing across `dev`/`train`/`triton`, so no change needed.

**Alternative:** `requirements_gns.txt` is only consumed by `setup.py` (unlike the others it's not a standalone `pip install -r` target), so the two pins could just be inlined into `extras_require` and the file dropped. But the new file is consistent with the repo's `read_requirements` pattern, so either way is reasonable.

Net: correct, low-risk, and a real improvement to the default install. Main asks: fix the contradictory description and keep the GNS True-branch covered in dev/CI.

gugarosa:
Thanks — addressed the two main asks in 57e90e7 + a description fix.

**1. Contradictory description (fixed).** Updated the top "What" line — it now says `dev` and `train` stay GNS-free too (only `[gns]` / `[train,gns]` / `[dev,gns]` pull the kernels), matching the code and the Notes section. Good catch; that was a leftover from the prior revision where dev/train did fold GNS in.

**2. GNS True-branch coverage.** Documented `pip install -e ".[dev,gns]"` in the README as the way to develop/test the Gram Newton-Schulz path, and noted that a plain `[dev]` install skips those cases (now a visible `importorskip` skip with a reason, not a silent pass). There's no CI workflow in the repo yet; when one's added, a single `[dev,gns]` job will keep the branch covered — flagging that as the follow-up.

**Minor (done).** Aligned the runtime `ImportError` to the extras form (`pip install -e ".[gns]"`, with the bare package as the fallback) so it matches the README's one consistent story.

**3 (behavior change) & 4 (upstream cutlass pin):** acknowledged — the install→runtime failure move is called out in the PR description's backward-compat note (no CHANGELOG in the repo to add a line to), and the only complete fix for GNS+FA4 coexistence is loosening `gram-newton-schulz`'s `nvidia-cutlass-dsl==4.4.2` to a 4.5.x range upstream. I'll track that as a separate follow-up; this PR unblocks the default install meanwhile.

JohnLangford:
This is John's reviewbot.

Solid, well-scoped change. I verified the core claims hold:

- **Correctness ✅** `gram_newton_schulz` is imported in exactly one place — the lazy, guarded branch in `megabatch_base.__init__` — and nothing in `dion/__init__.py` or the optimizer modules touches it at import time, so `from dion import ...` stays GNS-free. `quack-kernels` has no direct import anywhere (only transitive via `gram-newton-schulz`), so moving both together is right. The error-message branch is the only code change and is functionally a no-op.
- **Test coverage ✅** `test_dion2_post_ortho_triton.py` is the only test that exercises `use_gram_newton_schulz=True`, and the new `pytest.importorskip` is placed correctly before the optimizer is constructed. No other test imports the package at module scope.

A few substantive notes:

1. **Backward-compat (behavioral break, worth a release note).** Before this PR, `gram-newton-schulz`/`quack-kernels` were in `requirements_dion.txt` (i.e. `install_requires`), so *every* `pip install dion` / `dion[train]` / `dion[dev]` got them. Existing users who run with `use_gram_newton_schulz=True` on an unchanged install command will now hit the `ImportError` at runtime. That's the intended trade-off and the message is clear, but since `dion` is pre-1.0 and not on PyPI this silent change in default-installed deps deserves a CHANGELOG / release-note line so downstreams know to add `[gns]`.

2. **GNS path is now untested by default.** There's no `.github/workflows` in the repo today, but the consequence of `[dev]` no longer pulling the kernels is that a default `pip install -e .[dev] && pytest` run *skips* the GNS path entirely (importorskip → green). If/when CI is added, it must install `[dev,gns]` (on a cutlass-4.4.2-compatible image) or the GNS branch can rot silently behind the skip. Flagging so it's a conscious choice.

3. **Minor — runtime hint vs. install reality.** The new `ImportError` points at `pip install -e ".[gns]"`, which only works from a clone. A consumer who installed `dion` as a git dependency can't run `-e .`; the `(or pip install gram-newton-schulz)` fallback saves them, but for parity with the README you could also surface the `pip install "dion[gns] @ git+..."` form.

4. **Minor — explicit `quack-kernels==0.4.1` pin.** Since quack is only pulled transitively by `gram-newton-schulz==0.1.4`, the explicit line in `requirements_gns.txt` acts purely as a version pin. That's a reasonable reproducibility choice, but note it becomes a hard conflict if a future `gram-newton-schulz` requires a different quack version — worth a comment in the file so the coupling is obvious.

5. **Packaging nit (pre-existing class).** `setup.py` reads `requirements_gns.txt` at build time, like its siblings. None of these are in a MANIFEST.in, so an sdist build wouldn't bundle them — fine today because installs come from a full git checkout, just noting the new file inherits the same latent gap.

On elegance: the extra is the correct tool here — the `nvidia-cutlass-dsl==4.4.2` pin lives upstream in `gram-newton-schulz`, so dion can't relax it; isolating it behind an opt-in extra (and documenting that opting in re-introduces the FA4/Blackwell conflict) is the right scope. The only cleanup I'd consider is that every entry in `extras_require` repeats `install_requires + ...`, which setuptools already includes for extras — harmless and pre-existing, so no need to touch it in this PR. Consolidating the `requirements_*.txt` set into a `pyproject.toml` `[project.optional-dependencies]` would be the more modern shape, but that's a separate refactor and the file-based pattern here is internally consistent.

LGTM with the release-note (1) and CI-awareness (2) points addressed.

gugarosa:
Thanks — addressed both points in b5538ed.

**1. Release note (done).** Added a `CHANGELOG.md` with an `[Unreleased] → Changed` entry flagging the breaking install change: `gram-newton-schulz` / `quack-kernels` are no longer base/`dev`/`train` deps, downstreams running `use_gram_newton_schulz=True` need the `[gns]` extra, and the failure now surfaces as a runtime `ImportError` rather than silently-present kernels. (No CHANGELOG existed, so this also seeds the convention — Keep a Changelog style.)

**2. CI-awareness (acknowledged).** No `.github/workflows` in the repo today, so nothing to wire now. The README already documents `pip install -e ".[dev,gns]"` for the GNS path; when CI is added, a single `[dev,gns]` job on a cutlass-4.4.2-compatible image keeps the True-branch covered. Flagging as the conscious follow-up.

**Minor #4 (done).** Commented `requirements_gns.txt` to make the `quack-kernels ← gram-newton-schulz ← nvidia-cutlass-dsl==4.4.2` coupling explicit. `read_requirements` skips `#` lines, so the `[gns]` extra resolves unchanged.

**Minors #3 / #5:** the runtime hint keeps the `(or pip install gram-newton-schulz)` fallback (works for non-clone installs since it's on PyPI), and the `requirements_*.txt` / MANIFEST sdist gap is pre-existing and shared by the sibling files, so I left it out of scope here. The `pyproject.toml` `[project.optional-dependencies]` modernization is a reasonable separate refactor.

Thanks for the thorough passes.

NoahAmsel:
This looks sensible. It seems kind of unsustainable to actually manually update `requirements_gns.txt` every time cutlass is updated. We'll have to think about that...

JohnLangford:
This is John's reviewbot — follow-up on `b5538ed30` ("Add CHANGELOG release note … document quack coupling").

This closes out the two open items from my last pass:

- **Backward-compat (was point 1)** — the new `CHANGELOG.md` `[Unreleased] > Changed` entry is accurate and actionable: it flags the breaking install change, gives the exact `dion[gns]` opt-in (both git-extras and clone forms), and explains the runtime `ImportError` + the cutlass re-introduction caveat. 👍
- **quack coupling (was point 4)** — the `requirements_gns.txt` comment now makes the transitive `quack-kernels <- gram-newton-schulz==0.1.4 <- nvidia-cutlass-dsl==4.4.2` chain explicit and notes the explicit pin must track the gram-newton-schulz version. Exactly the intent.

Re-verified current HEAD: `requirements_dion.txt` is down to `numpy`/`torch`, the `gns`/`gram-newton-schulz` extras are wired, the lazy guarded import is the sole consumer, and the GNS test's `importorskip` is in place. No regressions.

Only residual (non-blocking, same as before): the runtime `ImportError` string still shows just `pip install -e ".[gns]"` + the standalone-package fallback, while the CHANGELOG/README now lead with the `pip install "dion[gns] @ git+…"` form. The fallback keeps it correct for non-clone installs, so this is cosmetic — optional to align.

LGTM. Ship it.

gugarosa:
Dug into the exact source of the pin — there is a concrete path here.

`requirements_gns.txt` does not set the cutlass version itself; the hard constraint is one upstream line:
- `gram-newton-schulz==0.1.4` → `nvidia-cutlass-dsl==4.4.2` (exact) + `quack-kernels>=0.4.1`
- `quack-kernels==0.4.1` → `nvidia-cutlass-dsl>=4.4.2` (already a floor)

So the only thing forcing the downgrade is gram-newton-schulz's exact `==4.4.2`, and it lives upstream in [`Dao-AILab/gram-newton-schulz`](https://github.com/Dao-AILab/gram-newton-schulz) — where their own sibling `quack-kernels` already uses `>=4.4.2`. The durable fix is a one-line relaxation there (`==4.4.2` → `>=4.4.2`), which we cannot merge ourselves but can flag.

Plan:
1. **Merge this PR as-is.** Isolating GNS behind `[gns]` is the correct downstream containment — base/`dev`/`train` stay cutlass-free, which is what unblocks FA4/Blackwell. The manual-bump burden is now scoped to opt-in GNS users only.
2. **File an issue on `Dao-AILab/gram-newton-schulz`** asking whether the `nvidia-cutlass-dsl` pin can be relaxed to a floor (matching their own `quack-kernels`).
3. **Cascade back here** if/when that lands upstream — float `requirements_gns.txt` (drop the exact pins) so `dion[gns]` auto-tracks whatever cutlass the image ships, and the manual update goes away entirely.

gugarosa:
Upstream follow-up filed: Dao-AILab/gram-newton-schulz#19 — asking whether `nvidia-cutlass-dsl==4.4.2` can be relaxed to a floor (their own `quack-kernels` already uses `>=4.4.2`). If that lands upstream, we cascade back here by floating `requirements_gns.txt` so `dion[gns]` auto-tracks the image cutlass and the manual bump disappears.

PR #94, opened 2026-06-22
Fix AdamW LM head fallback scaling

jibby1729:
Currently, the LM output head scaling for Adam fallback is `$1/\sqrt{d_\text{in}}$`, but that should only be applied to Lion fallback.

Adam has adaptive scaling and approximately ensures RMS 1 updates. Lion doesn't. Empirically, scaling by `$1/\sqrt{d_\text{in}}$` for the output head with Adam fallback leads to worse perf than it does with keeping it 1.

It makes sense to have this scaling with Lion, and the derivation for it is in Appendix D.2 of the original Dion paper: https://arxiv.org/pdf/2504.05295

In Figure 14 of the paper, it even says in the caption, "Lion is tested with learning rates scaled by `$1/\sqrt{d}$`, `$1/d$`, or left unscaled, while Adam uses a single fixed rate for all non-matrix parameters." This implies the original paper did not have this scaling with the AdamW fallback.

## Changes

- Use the base learning rate for LM head params when `scalar_opt == "adamw"`
- Keep `$1/\sqrt{d_\text{in}}$` scaling when `scalar_opt == "lion"`
- Update README examples and the parameter-group table to match this behavior

## Validation

- `PYTHONPYCACHEPREFIX=/tmp/dion_pycache python -m py_compile train.py`
- `tests/test_optimizers.py::TestMixedParamGroups`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ All notable changes to this project are documented in this file.
 
 ### Changed
 
+- AdamW scalar fallback now uses the base learning rate for LM head parameters,
+  while Lion fallback keeps the `1 / sqrt(d_in)` LM-head scaling. This affects
+  shipped `configs/*_160m.yaml` runs, which set `scalar_opt: adamw`.
+
 - **Breaking (install):** `gram-newton-schulz` and `quack-kernels` are no longer
   base dependencies. They moved to an optional `dion[gram-newton-schulz]` extra
   (alias `dion[gns]`), and are also excluded from the `dev` and `train` extras.
--- a/README.md
+++ b/README.md
@@ -183,7 +183,7 @@ The details of parameter grouping are dependent on model architecture and implem
 * Biases in `nn.Linear` layers (if used) are one-dimensional vectors, which must be placed into a separate parameter group from the weight matrices. Use Lion or AdamW.
 * Normalization layers (e.g. `nn.LayerNorm`, `nn.RMSNorm`) may contain vectors of learnable weights. Use Lion or AdamW.
 * Embedding layers (e.g. `nn.Embedding`) are stored as 2D tensors, but should be treated as a collection of 1D vectors using Lion or AdamW. (Warning: using Dion here will run without error, but will give poor performance.)
-* Unembedding layers (e.g. LM head) are typically implemented as a `nn.Linear` layer, but shoud also be treated as a collection of 1D vectors. Furthermore, they should use a **smaller scaled learning rate**. It is very important to manually identify this layer and place it into its own parameter group, as it is otherwise indistinguishable from weight matrices!
+* Unembedding layers (e.g. LM head) are typically implemented as a `nn.Linear` layer, but shoud also be treated as a collection of 1D vectors. When using Lion as the scalar optimizer, they should use a **smaller scaled learning rate**; with AdamW, use the base learning rate unless you intentionally tune a separate AdamW learning rate. It is very important to manually identify this layer and place it into its own parameter group, as it is otherwise indistinguishable from weight matrices!
 (Warning: using Dion here will run without error, but will give poor performance.)
 * Convolution layers typically use parameter tensors with 3+ dimensions. These are currently not supported for Dion. Support for convolution layers in Muon is experimental, and can be enabled with the option `flatten=True` to automatically flatten them to 2D matrices when computing the optimizer update.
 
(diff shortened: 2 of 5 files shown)


JohnLangford:
This is John's reviewbot.

**Verdict:** correct and well-motivated. AdamW's `m/sqrt(v)` normalization already yields ~unit-RMS updates, so the extra `1/sqrt(d_in)` factor double-scales the head LR (at `d=768` it shrinks it ~28x); Lion's sign update is the case the App. D.2 derivation actually targets. The conditional preserves Lion, `scalar_opt` is already validated to `{adamw, lion}` just above so the ternary is exhaustive, `math` is imported, and the AdamW head group now cleanly mirrors the embedding group (base `lr`, same betas, `wd=0`). README table/prose are accurate.

A few substantive points:

**1. This is not a no-op for the shipped configs — please surface it.** All four `configs/*_160m.yaml` set `scalar_opt: adamw`, so this raises the lm_head LR from `0.02/sqrt(768) ≈ 7.2e-4` to `0.02` (~28x) for the repo's default runs and breaks comparability with any prior curves/checkpoints. That's the intent, but it warrants (a) a `CHANGELOG.md` `[Unreleased] → Changed` entry (the repo maintains one), and (b) ideally a link to the loss curve / run behind "empirically ... worse perf" — that evidence is the load-bearing justification for changing default behavior, and the Lion-vs-Adam asymmetry isn't obvious from first principles since both produce ~unit-RMS updates.

**2. The cited test doesn't exercise the changed code.** `TestMixedParamGroups` builds param groups by hand and calls the optimizer classes directly — it never imports `train.init_optimizer`, never sets `scalar_opt`, and never asserts on the head LR. It passes with or without this patch, so it isn't evidence the fix works. There's currently no test pinning the per-fallback head LR. A small assertion — `init_optimizer` yields an lm_head group with `lr == lr` for adamw and `lr/sqrt(d)` for lion — would lock the behavior and catch a future regression. The group-assembly logic is pure-Python and testable on CPU, even though `test_optimizers.py` is CUDA-gated.

**3. Optional (elegance).** The scaling is a property of the *fallback optimizer*, not the layer. Factoring a tiny `head_lr_scale(scalar_opt)` (→ `1/sqrt(d)` for lion else `1.0`) makes it unit-testable without constructing a model and states the intent in one place; a one-line comment pointing at Dion paper App. D.2 / Fig. 14 would help future readers.

No performance/throughput/security impact — this is construction-time hyperparameter wiring.

JohnLangford:
This is John's reviewbot.

Re-reviewed after `7f1164c` ("Document and test LM head LR scaling") — this cleanly resolves all three points from my prior review:

- **CHANGELOG** entry added under `[Unreleased] → Changed`, and it explicitly calls out that the shipped `configs/*_160m.yaml` (all `scalar_opt: adamw`) are affected. 👍
- **Factored helper** `lm_head_lr_scale(scalar_opt, model_dim)` in `dion/opt_utils.py` with a comment pointing at App. D.2 / Fig. 14 — exactly the elegance fix, and `init_optimizer` now reads as `hp.lr * lm_head_lr_scale(...)`.
- **Real test** `test_lm_head_lr_scale_by_scalar_opt`, parametrized over adamw/lion, CPU-only (no CUDA gate) so it actually runs in most environments.

Also confirmed the `import math` move is safe — there are no remaining `math.` references in `train.py`, so dropping the import there won't `NameError`.

Two minor, non-blocking residuals:

1. The test pins the helper (good), but nothing asserts `init_optimizer` actually applies the scale to the *lm_head* group (vs. embedding). Low priority given the wiring is now a trivial one-liner — optional.
2. `lm_head_lr_scale` silently returns `1.0` for any non-`"lion"` string. Safe today because `init_optimizer` validates `scalar_opt ∈ {adamw, lion}` upstream, but as a public util a typo (`"adam"`) would pass unnoticed. A guard/assert is optional.

And the only still-open item from before: a pointer to the loss curve / run behind the "empirically worse perf" claim would be nice to capture in the PR, since it's the justification for changing default behavior — but not a blocker.

LGTM. Nice turnaround.

jibby1729:
🤖 edit made using GPT 5.5x high on Codex. Following up on @JohnLangford's non-blocking request for a pointer to the run behind the empirical claim.

I ran a small controlled modded-nanoGPT validation using the minimal Muon-introduction snapshot, `kellerjordan/modded-nanogpt@c26c2a1` (`new optimizer`, Oct 4 2024). I replaced the local Muon optimizer with local `DionSimple`, kept AdamW as the fallback for the tied `lm_head` / embedding weight, and held the settings fixed except for the AdamW fallback head scale.

Validation loss curves

Run setup:

- Model/data: `d12`, FineWeb GPT-2 token shards from `kjj0/fineweb10B-gpt2`
- Hardware: 4x B200 per run
- Training: 500 steps, validation every 50 steps, seed `1337`
- Correct run: AdamW fallback head scale `1.0`, SLURM job `117993`
- Incorrect run: AdamW fallback head scale `1 / sqrt(768) = 0.03608439`, SLURM job `117994`

Trimmed logs, starting at `step 1/500`:

- Correct scale log: https://raw.githubusercontent.com/jibby1729/dion/9dc6559333ecaa79d0f4fa4516a394427e1b95fa/pr-94-validation/dion_correct_500_slurm-117993.out
- Incorrect scale log: https://raw.githubusercontent.com/jibby1729/dion/9dc6559333ecaa79d0f4fa4516a394427e1b95fa/pr-94-validation/dion_incorrect_500_slurm-117994.out

Validation loss:

| Step | AdamW head scale `1.0` | AdamW head scale `1 / sqrt(768)` |
|---:|---:|---:|
| 0 | 11.001800 | 11.001800 |
| 50 | 6.296978 | 7.941091 |
| 100 | 5.668626 | 6.737638 |
| 150 | 5.227561 | 6.235789 |
| 200 | 4.951087 | 5.907558 |
| 250 | 4.714558 | 5.666320 |
| 300 | 4.521734 | 5.468090 |
| 350 | 4.415174 | 5.290514 |
| 400 | 4.334936 | 5.149511 |
| 450 | 4.272649 | 5.035650 |
| 500 | 4.225253 | 4.942618 |

The two runs had essentially the same throughput and peak memory, around `556k tok/s` and `63084 MiB`. At 500 steps, the `1 / sqrt(768)` AdamW fallback scale is `+0.717` validation loss worse than keeping the AdamW fallback head scale at `1.0`.

jibby1729:
Set up a quick experiment with Codex on [modded-nanogpt](https://github.com/kellerjordan/modded-nanogpt) to produce these plots. I originally noticed this issue at work (can't post those plots) but the discrepancy is quite noticeable. 

Will sign the CLA after I get perms

jibby1729:
@microsoft-github-policy-service agree company="Block"

jibby1729:
Signed the CLA and put my commit on my work email, ready to merge

JohnLangford:
Thanks for this.  I've been conducting my own tests---the issue is real but was hidden because we were using tied embeddings.   I'm running one last test now.

JohnLangford:
Merged, thank you :-)

Issue #96, opened 2026-06-30
Dion2/NorDion2 FSDP2 padding regression


alint77:
## Problem

Dion's whole point is a **cheaper orthogonalization step**: select only a
`fraction` (default 0.25) of each matrix's rows and run Newton-Schulz on that
submatrix. Under FSDP2 row-sharding this saving is silently **erased**.

In `megabatch_orthogonalize_async` (`dion/megabatch_base.py`), every rank's
selected rows are padded back up to the full per-rank shard before the
all-to-all:

```
padded_local_size = ceil(global_comm_dim_size / world_size)   # the FULL shard
# e.g. 5120-row weight, 4 ranks: selected 320 rows -> padded back to 1280
```

Because `select_dim == shard_dim` (both -2/rows), the selection dimension is
exactly the dimension the pad rule re-inflates. Result: the assembled matrix is
full size (~¾ zero rows), so **comm volume and NS GEMM cost are identical to
NorMuon** — Dion2's `fraction` buys nothing here.

Not a correctness bug (zero rows contribute nothing to UᵀU and are narrowed off
after the second all-to-all). It is a **performance regression**: Dion2 pays the
selection overhead (topk, gather, scatter, ~244-kernel tail) and collects none of
its compute/comm reward.

## How it was identified

- GPU kernel breakdown of `nordion2_gramns` vs `normuon_gramns` (5120/8L, bs1
  traces): both spend ~63% in the same gramNS GEMMs; nordion2's NS is only
  modestly cheaper (49 vs 65 ms), not the ~4x expected from `fraction=0.25`.
- Trace shapes confirmed it: nordion2's `all_to_all` chunks and
  `gemm_symmetric` inputs are **full `[*, 5120, 5120]`**, byte-identical comm
  (2500 MB / 6 calls) to normuon — the selection never shrank the matrix.
- `git blame` + history traced the padding to the megabatch path and showed the
  original Dion2 did **not** do this.

## When it was introduced

The original Dion2 (Kwangjun Ahn) had a **dedicated** distributed path that
communicated and orthogonalized **only the selected submatrix** — the old
`dion2_update_batch_async` sized its all-to-all buffers to `U_selected` and
carried the literal comment `# Only submatrix is orthogonalized!`. No pad to
global size.

The regression was introduced by the **megabatch refactor** that moved Dion2
onto NorMuon's shared comm path, then hardened by later padding fixes:

| date | commit | what |
|---|---|---|
[...]

PR #97, opened 2026-07-01
Add selection_scope to Dion2/NorDion2 (fix FSDP2 padding regression)

alint77:
Relates to #96.

On the FSDP2 row-sharded path, Dion2/NorDion2 selected a `fraction` of rows, but the megabatch all-to-all padded each shard's selection back to full local size before communication — erasing the saving (Newton-Schulz ran on the full matrix, comm matched NorMuon), so `fraction < 1` cost as much as `fraction = 1`.

Adds a `selection_scope` option (default `"local"`):

- **`"local"`** — uniform `k = ceil(fraction * padded_local)` per rank, so alltoall sizes match and padding is skipped; comm and Newton-Schulz shrink by `fraction`. Numerically identical to the old behavior. Selected set is world-size variant.
- **`"global"`** — communicate the full shard, take top-k on the assembled matrix; exact and invariant to sharding layout, at full comm cost.

Off the row-sharded path selection_scope is a no-op. Adds `tests/test_dion2_selection_scope.py`.

### DIFF
--- a/dion/dion2.py
+++ b/dion/dion2.py
@@ -329,23 +329,22 @@ def dion2_update_megabatch_async(
         return
 
     # --- Local selection (default): per-shard top-k, communicate only the
-    # selected rows. On the row-sharded path, when the global comm dim divides
-    # evenly across ranks every rank holds exactly ``padded_local`` rows, so a
-    # uniform ``k = ceil(fraction * padded_local)`` is selectable on every rank
-    # and never exceeds the local size. We then tell the megabatch the global
-    # size is ``k * world_size``: it computes padded_local_size = k, sees
-    # U_selected already at that size, and skips padding entirely -- comm and
-    # Newton-Schulz both shrink by ``fraction``. When the dim does not divide
-    # evenly (uneven/empty shards, e.g. a tiny num_heads dim), fall back to the
-    # original full-shard padding so the alltoall sizes still match; this is the
-    # pre-existing behavior and stays numerically identical.
-    if comm_dim is not None and global_dim_size % world_size == 0:
-        padded_local = global_dim_size // world_size
+    # selected rows. Under FSDP2 contiguous chunking every rank holds at most
+    # ``padded_local = ceil(global / world_size)`` rows, so a uniform
+    # ``k = ceil(fraction * padded_local)`` is the per-rank selected count. We
+    # select up to ``k`` rows locally (short/empty shards select fewer -- see
+    # dion2_pre_orthogonalize) and tell the megabatch to pad every shard to
+    # exactly ``k`` via ``local_comm_size=k``, so the alltoall stays uniform
+    # while comm and Newton-Schulz both shrink by ``fraction``. This holds for
+    # uneven divisions too (the remainder/empty ranks just contribute zero-padded
+    # rows), so there is no even-division special case. ``global_comm_dim_size``
+    # keeps its true meaning (the unsharded size).
+    if comm_dim is not None:
+        padded_local = (global_dim_size + world_size - 1) // world_size
         k = max(1, int(math.ceil(fraction * padded_local)))
-        global_comm_dim_size = k * world_size
     else:
         k = None
-        global_comm_dim_size = global_dim_size
+    global_comm_dim_size = global_dim_size
 
     # Pre-orthogonalize: momentum update + submatrix selection
     U_selected, indices_list = dion2_pre_orthogonalize(
@@ -368,6 +367,7 @@ def dion2_update_megabatch_async(
         flatten=flatten,
         epsilon=epsilon,
         global_comm_dim_size=global_comm_dim_size,
+        local_comm_size=k,
     )
 
     # Compute scaled learning rate
@@ -444,8 +444,9 @@ def dion2_pre_orthogonalize(
     "local" path so every rank selects the same count, derived from the global
     size, instead of ``ceil(fraction * local_size)``). A rank whose local shard
     is shorter than ``k_override`` selects all of its rows; ``topk`` is clamped
-    to the available count and the result is zero-padded up to ``k_override`` so
-    the downstream alltoall sees a uniform per-rank size.
+    to the available count. The pad up to ``k_override`` is not done here -- the
+    downstream megabatch pads U to ``local_comm_size=k_override`` for a uniform
+    alltoall, and indices stay at the real selected count.
     """
     dtype = M[0].dtype
 
@@ -453,10 +454,20 @@ def dion2_pre_orthogonalize(
     # select_dim is the dimension we select submatrix from
     num_select = M[0].size(select_dim)
     norm_dim = -1 if select_dim == -2 else -2
+    # k is the requested selected count; k_topk is what topk can actually take
+    # from this shard (clamped to its real rows). When the shard is shorter than
+    # k_override -- possible on the last/remainder rank under uneven FSDP2
+    # chunking -- we select all its rows here. U_selected then carries only the
+    # real k_topk rows; the pad up to k happens downstream in
+    # megabatch_orthogonalize_async (local_comm_size=k) purely so the alltoall
+    # sees a uniform per-rank size. indices deliberately stay at k_topk: the
+    # megabatch narrows its result back to k_topk before it is scattered, so the
+    # padded rows never reach post_orthogonalize.
     if k_override is not None:
         k = k_override
     else:
         k = max(1, int(math.ceil(fraction * num_select)))
+    k_topk = min(k, num_select)
 
     # Update momentum: M = M + G
     G = [g.to(dtype=dtype) for g in G]
@@ -482,8 +493,9 @@ def dion2_pre_orthogonalize(
     # Compute L1 norm along norm_dim (sum of absolute values)
     slice_norms = M_stacked.norm(p=1, dim=norm_dim)
 
-    # Batched topk: indices shape (batch_size, k)
-    _, indices = torch.topk(slice_norms, k, dim=-1, sorted=False)
+    # Batched topk: indices shape (batch_size, k_topk). k_topk <= num_select is
+    # guaranteed, so this never raises even on a short remainder shard.
+    _, indices = torch.topk(slice_norms, k_topk, dim=-1, sorted=False)
 
     # Extract the selected rows/columns from each momentum tensor.
     # `indices` has shape (..., k) where k is the number of selected slices.
--- a/dion/dion2.py
+++ b/dion/dion2.py
@@ -65,12 +65,17 @@ class Dion2(DistributedOrthoBase):
         newton_schulz_func: Optional[Callable] = None,
         verbose: bool = False,
         triton_post_ortho: bool = False,
+        selection_scope: str = "local",
     ):
         # Validate hyperparameters
         if lr < 0.0:
             raise ValueError(f"Invalid learning rate: {lr}")
         if not (0.0 < fraction <= 1.0):
             raise ValueError(f"fraction must be in (0, 1], got {fraction}")
+        if selection_scope not in ("local", "global"):
+            raise ValueError(
+                f"selection_scope must be 'local' or 'global', got {selection_scope!r}"
+            )
         if ef_decay < 0.0:
             raise ValueError(f"Invalid ef_decay: {ef_decay}")
         if len(betas) != 2 or betas[0] < 0.0 or betas[1] < 0.0:
@@ -92,6 +97,7 @@ class Dion2(DistributedOrthoBase):
             adjust_lr=adjust_lr,
             algorithm="dion2",
             step=0,
+            selection_scope=selection_scope,
         )
         super().__init__(
             params, distributed_mesh, "dion2", defaults,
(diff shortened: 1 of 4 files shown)


alint77:
@NoahAmsel @JohnLangford. Please test the fix for numerical accuracy and loss curve, I had limited compute so couldn't do it myself.

JohnLangford:
This is John's reviewbot.

Solid fix for a real regression. I traced both scopes through `megabatch_orthogonalize_async` and the math holds: dropping the padded zero rows doesn't change `Xᵀ X` (the local top-k of `k` rows and the old full-pad path produce the same orthogonalization on the nonzero rows), and the global path's assemble-then-select is exact and world-size invariant. The NorDion2 masked normalization is also correct — Frobenius norm and per-neuron variance are invariant to zero rows, and the `where(sel, …)` mask discards the spurious V/momentum updates on non-selected rows. Nice.

A few substantive points:

**1. "Numerically identical" is exact-arithmetic only, not bit-exact.** The default flips from "pad to full, NS on `global` rows" to "NS on `k·world_size` rows." `Xᵀ X` is mathematically identical, but the reduction runs over a different number of rows, so GPU accumulation order differs → results match only up to fp noise (~1e-7), which is why the test uses `atol=2e-3`. Worth stating explicitly in the PR: **BPB/loss will match pre-PR within noise, not byte-identically.** Anyone validating this fix by expecting a byte-identical BPB (the usual signal here) will be misled.

**2. `k_override` docstring contradicts the code.** The docstring promises it "clamps `topk` to the available count and zero-pads up to `k_override`" for short shards. The code does neither — `torch.topk(slice_norms, k_override)` raises if `k_override > num_select`. It's safe *only* because the caller gates on `global_dim_size % world_size == 0` (⇒ every shard is exactly `padded_local`, so `k ≤ num_select`). Either implement the clamp or change the docstring to state the precondition. As written it's a latent trap if `k_override` is ever reused off that guarded path.

**3. The default "local" fix is only smoke-tested at `fraction < 1`.** `test_local_scope_runs_and_updates_selected` just asserts the weights moved; `test_fraction_one_local_equals_global` runs at `fraction=1.0` where there's no selection at all. So the headline claim — "local at `fraction<1` is numerically identical to the old full-pad path" — has no direct assertion. That's the exact behavior justifying making `local` the default; it deserves a reference comparison (e.g. local-scope 2-rank vs a single-process per-shard top-k reference at `fraction=0.25`). Also missing: the uneven-division fallback (`global % world_size != 0`), the unsharded no-op case, and the `ValueError` on a bad scope string. And the whole file is `skipif CUDA<2`, so none of it runs in CPU CI.

**4. Uneven/tiny shards silently keep the regression.** When `global_dim_size % world_size != 0`, "local" falls back to full-shard padding — full comm, no saving. It's commented, but there's no `verbose` note, so a user with such params gets none of the win and no signal. Cheap generalization: `all_gather` the per-rank `k` (one tiny int collective) and pad to `max_k` instead of `padded_local` — recovers most of the saving for uneven shards and removes the even-division special-case entirely.

**5. Overloading `global_comm_dim_size` is a bit of a hack.** Passing `k*world_size` as "the unsharded global size" so the megabatch back-computes `padded_local_size = k` and skips padding is clever but repurposes a parameter documented as the true global size, and correctness silently depends on the internal `padded_local_size < original_local_size` check lining up. An explicit `skip_pad=True` / `local_comm_size=k` arg on `megabatch_orthogonalize_async` would be more honest and self-documenting.

**6. Design / minor.** The global+local scaffolding (scope decision, even-division `k`, the `adjust_lr` dispatch) is now duplicated near-verbatim across `dion2_update_megabatch_async` and `nordion2_update_megabatch_async` — worth a shared helper to avoid drift. `_make_select_and_orthogonalize` returns an uncompiled closure, so its topk/gather/scatter run eager each step (acceptable since global is the expensive path, but note it). And nordion2 now imports the private `_normuon_normalization_core` across the module boundary.

None of these block the core fix — the default local path is correct and does what #96 needs. Items 1–3 are the ones I'd want addressed before merge (set the BPB expectation, fix the `k_override` doc, and add a real numerical-identity test for the default path).

alint77:
the throughput/timing/scaling/comms results are in #96

JohnLangford:
This is John's reviewbot. Re-review after `4f52709` (thanks for the fast iteration — it cleanly resolves my earlier #2/#4/#5).

**The new commit is correct.** I traced the uneven-shard path: `local_comm_size=k` pads each shard to `k` while `global_comm_dim_size` keeps its true meaning (good — correctness no longer rides on an incidental internal check); `k_topk=min(k,num_select)` keeps `topk` in range on short/remainder ranks; real rows stay first, indices stay at the real count, and the megabatch narrows back to `k_topk` before scatter so zero-pad rows never reach the post step. Empty shards (`num_select==0`) short-circuit and no-op. The `local_comm_size + split_sizes` rejection is the right call.

Two things still worth addressing:

**1. Unflagged semantics change on uneven shards.** Dropping the even-division case means ragged layouts now select a *uniform* `k = ceil(fraction·ceil(global/W))` per rank (clamped), not the pre-PR *proportional* `ceil(fraction·local_size)`. On a short shard that selects **more** rows than before (e.g. fraction .25, shard=4, max=8 → new picks 2, old picked 1), so total work is ~`fraction·padded_global` rather than `fraction·global`. This is defensible (more uniform), but it means "numerically identical to prior behavior" now holds **only for even divisions** — please scope that claim in the docstring/PR text.

**2. The new path has zero test coverage.** `test_dion2_selection_scope.py` is unchanged (world_size=2, 16 rows → always even), so the uniform-k uneven/empty-shard logic — exactly where alltoall-size and narrow-offset bugs hide — is untested. Add a remainder case (e.g. 18 rows over W=8 so ranks 6–7 are short/empty, or W=2 with odd rows) asserting no-hang + local-scope correctness.

**Cluster evidence** (I ran the A/B: Dion2, 1.5B, FSDP2 shard-8, patch vs #90-base, same seed/data, n=32 steady-state steps):
- **No regression:** arms converge together (local 12.34 vs base 12.35), no NaN, and the global/masked-post `@torch.compile` path runs clean at scale.
- **Throughput (isolated optimizer step):** padded-base `0.0591±.0005` → local `0.0551±.0012` (−6.8%) → global `0.0506±.0019` (−14%). Clean and real, but only **~1–2% end-to-end** — the Dion step is ~15% of step time and the megabatch all-to-all overlaps. Peak memory identical (82.7 GB; comm buffers aren't the bottleneck).

**Design flag (the surprise):** `global` is the **fastest** scope here, not the most expensive. local and global had identical fwd+bwd (same nodes) yet global's optimizer step was lower — it skips the per-rank `topk`/`gather`/EF-scatter machinery that `local` runs on every rank, and its extra comm is hidden by overlap. So the *cost* rationale for making `local` the default doesn't hold at production scale, and `local` also sacrifices world-size reproducibility (its selected set is layout-variant). I'd reconsider defaulting to `global` (exact + layout-invariant + reproducible + empirically faster here), or at minimum document that `global` isn't more expensive — pending a comm-bound / large-shard datapoint where `local`'s smaller alltoall might finally win.

Backward-compat/security are fine: `local_comm_size` is opt-in (default `None`), so Muon/NorMuon and every existing caller are untouched; no new surface.

JohnLangford:
This is John's reviewbot — follow-up to the throughput note above, now with a second scale point that turns the "global is faster" observation into a trend.

Same A/B (Dion2, FSDP2 shard-8, fine-sampled, fwd+bwd matched across arms so the optimizer step is isolated), padded-#90-base vs the two scopes, at two model sizes:

| scope | 1.5B opt step (n=32) | 3B opt step (n=63) |
|---|---|---|
| padded (#90 base) | 0.0591 | 0.0651 |
| **local** | 0.0551 (−6.8%) | 0.0598 (−8.2%) |
| **global** | 0.0506 (−14.4%) | 0.0578 (−11.3%) |

The trend across scale: **local's saving grows** (6.8% → 8.2%) while **global's lead over local shrinks** (7.6pp → 3.1pp). That's the crossover dynamic I flagged as "pending a comm-bound datapoint" — as matrices grow the Dion step gets comm-heavier, so `local`'s ~4× smaller all-to-all increasingly pays off and erodes `global`'s fixed selection-overhead advantage.

So the default question is genuinely scale-dependent, not a one-way "global is cheaper":
- At **1.5–3B** (what I measured), `global` is faster *and* exact/layout-invariant → the better default here.
- Extrapolating the trend, `local` overtakes at **larger matrices / comm-bound (inter-node shard)** regimes → the PR's choice of `local` as default is defensible there.

Either way the end-to-end delta is ~2–3% of step time (the Dion step is ~15% and the all-to-all overlaps), so this is a tie-breaker, not a headline. My suggestion stands: pick the default deliberately with a one-line comment on the reasoning (and note `local` trades away world-size reproducibility for that comm win), rather than defaulting to `local` purely on an assumed cost advantage that doesn't hold at the scales tested.

JohnLangford:
This is John's reviewbot — closing the throughput thread with a third scale point (native 32-GPU / 4-node, mbs-4) that completes the crossover trend and **flips my earlier "reconsider defaulting to global" into an endorsement of the `local` default.**

Optimizer step vs the padded #90 baseline, same A/B, fwd+bwd matched across arms:

| scale / config | local | global | local↔global gap |
|---|---|---|---|
| 1.5B, 8-GPU | −6.8% | −14.4% | 7.6 pp |
| 3B, 8-GPU | −8.2% | −11.3% | 3.1 pp |
| 3B, 32-GPU / 4-node | −8.7% | −10.1% | **1.4 pp** |

The gap closes monotonically (7.6 → 3.1 → 1.4 pp): as matrices grow and the Dion step gets comm-heavier, `local`'s ~4× smaller all-to-all increasingly pays off, erasing `global`'s fixed selection-overhead advantage. By 3B/32-GPU the two are within noise, and extrapolating one model size `local` wins outright. So `local` as the default is the right call for production-scale models — it converges to global's speed while keeping the comm saving that dominates at scale (trading away only world-size reproducibility, which is the honest cost to document).

Two supporting facts from the multi-node run:
- **Robustness:** all three scopes ran clean across 4 nodes at native mbs-4, no hang/OOM/NaN; peak memory identical across scopes at every scale (82.7 GB @1.5B, 117.7 GB @3B-mbs4) — the scope choice doesn't move peak memory.
- **The patched path is intra-node:** 32-GPU `opt_s` ≈ the 8-GPU shard-8 values, because the 4-node replicate dimension is DDP gradient all-reduce; the Dion all-to-all never leaves the shard group. So multi-node adds nothing to the patched path (measured, not assumed).

Net recommendation unchanged on correctness (LGTM once the uneven-shard path gets a test + the "numerically identical" claim is scoped to even divisions), and now positively: keep `local` as default, and a one-line comment noting it's a deliberate scale-aware choice (comm-saving that wins at large scale, at the cost of layout-variant selection) would make the rationale self-documenting.

JohnLangford:
This is John's reviewbot — **retracting my earlier "keep `local` as default" addendum** and recommending the opposite, based on a convergence A/B I should have run before making that call.

**Data:** `local` vs `global`, same pin/config/seed/data, 100M tokens (381 steps), constant LR after warmup, 2 seeds each (1.5B dense, FSDP2 shard-8):

| scope | ntp_loss (last step) |
|---|---|
| global s42 / s123 | 3.861 / 3.864 |
| local  s42 / s123 | 3.937 / 3.971 |

`global` converges **~0.09 nat lower**, and the seed spread (global 0.003, local 0.034) is far smaller than the gap — so it's a real effect, reproduced in both seeds, not noise. `local`'s per-shard selection is a sharding-dependent *approximation* of the true top-k, and here that approximation costs measurable quality. (Caveat: early-training / constant-LR; a longer run could shift the magnitude, but it already refutes "local is a free win.")

Combined with the throughput results (global was faster-or-tied at every scale tested; local only closes the gap by 3B/32-GPU), **`global` wins on every axis I measured**: exact, world-size-reproducible, faster-or-tied, *and* better-converging. So the recommendation is:

> **Make `global` the default; keep `local` as an opt-in** for the comm-bound large-scale regime where its cheaper all-to-all may eventually pay off.

Concrete change (both optimizers; `local` stays valid, just no longer the default):

```diff
-        selection_scope: str = "local",
+        selection_scope: str = "global",
```
in `Dion2.__init__` and `NorDion2.__init__` (+ the mirrored defaults on `dion2_update_megabatch_async` / `nordion2_update_megabatch_async`), plus docstring/comment updates flipping the "(default)" annotation and documenting the tradeoff. The `selection_scope not in ("local","global")` validation is unchanged, so `local` remains fully supported. Since this PR *introduces* the option, there are no existing users relying on the old default — flipping it now is clean. I have the full patch staged and can open it against your branch if useful.

Everything else from my prior review still stands (add a test for the uneven/empty-shard path; scope the "numerically identical" claim to even divisions).

JohnLangford:
This is John's reviewbot. Opened alint77/dion#1 against this branch with the two changes from the review + cluster A/B: (1) default `selection_scope` to `"global"` (keeps `"local"` as opt-in) — global converged ~0.09 nat lower in a matched-seed 1.5B A/B and is faster-or-tied on opt-step at every scale tested; (2) the `selection_scope` test fix (wide matrix + bf16 tolerance + a `no-pad==pad` check) → 8/8 on 2 GPUs. Merging that PR folds both into this branch. Still open: an uneven/empty-shard test for the new local generalization, and scoping the "numerically identical" wording to even divisions.

PR #104, opened 2026-07-16
Make the megabatched optimizer step CUDA-graph-capturable

JohnLangford:
## What

Adds `dion.cuda_graph.CudaGraphOptimizer`, a drop-in wrapper that captures `optimizer.step()` into a CUDA graph and replays it. The megabatched Muon/NorMuon/Dion2/NorDion2 step issues hundreds of small host ops per step for its per-matrix selection / error-feedback / normalization bookkeeping; on a sharded step that dispatch is a fixed per-step host cost that doesn't shrink with the data-parallel group. The step is fixed-shape (top-k selection changes which rows are chosen, not the shapes), so one capture + replay collapses the whole host dispatch — including the megabatch all-to-all (NCCL) and the symmetric-GEMM kernels — into a single graph launch, with **GPU (device) time unchanged**.

```python
opt = CudaGraphOptimizer(NorDion2(param_groups, ...), warmup_steps=10)
for batch in loader:
    loss = model(batch); loss.backward()
    opt.step()                          # eager for warmup_steps, then captured + replayed
    opt.zero_grad(set_to_none=False)    # keep .grad buffers stable for replay
```

## Making the step capturable

- **Scheduled LR.** Each param group carries its learning rate as a persistent device tensor (`megabatch_base._group_lr_tensor`), refreshed in place from the (scheduler-updated) python `group["lr"]` each step **outside** the graph (skipped while capturing; the wrapper refreshes before each replay). So a per-step LR schedule takes effect under replay instead of freezing at the capture-time value.
- **Capturable AdamW.** `adamw_update_foreach` gains a capturable form: per-param device step tensors incremented on-device, with `lr` read through the fused kernel's `tensor_lr` overload (`grad_scale=None, found_inf=None` select it — omitting them silently resolves to a different overload that misreads the step). The legacy `step: int` form (baked step + float lr) is **unchanged**, so existing callers — including the non-megabatched `Dion` — and `tests/test_adamw_foreach.py` are untouched.
- **Post-capture replay.** `CudaGraphOptimizer._capture` replays once immediately after capture: CUDA-graph capture *records* ops without executing them, so without this the captured step runs one step short.

## Numerics

`graph == eager` is **bit-exact** for the new code (both use the device-tensor LR), for a constant *and* a per-step-scheduled LR. Moving the matrix-path LR from a CPU wrapped-scalar to a device tensor changes bf16 type promotion in the update multiply (`neg_lr * U`), a ~1.5e-5-relative (bf16-rounding-class) shift versus the previous CPU-scalar path — below the optimizer's inherent bf16 noise, and it does not affect the AdamW path or the run-to-run determinism / parity tests.

## Testing

New `tests/test_cuda_graph.py`: capture-vs-eager and scheduled-LR-under-replay, for Muon/NorMuon/Dion2/NorDion2 (matrix + AdamW-scalar groups together) — 8 tests. Existing suites pass unchanged on the affected paths: `test_adamw_foreach` (13), `test_optimizers` (89), `test_dion2_post_ortho_triton` / `test_state_prepopulation` / `test_newton_shulz` / `test_polar_express` (114 passed, 16 skipped). The multi-GPU `test_dion2_selection_scope` / `test_megabatch_empty_shard` need a multi-GPU runner (CI).

### DIFF
--- a/dion/cuda_graph.py
+++ b/dion/cuda_graph.py
@@ -12,19 +12,30 @@ graph launch (GPU time unchanged), and the capture holds through the megabatch a
 Usage (drop-in around any optimizer whose step() is graph-safe -- no host syncs, fixed
 shapes, gradients living in stable ``.grad`` tensors):
 
-    opt = CudaGraphOptimizer(Dion2(param_groups, ...), warmup_steps=10)
+    inner = Dion2(param_groups, ...)
+    sched = torch.optim.lr_scheduler.CosineAnnealingLR(inner, T_max=...)  # NOT the wrapper
+    opt = CudaGraphOptimizer(inner, warmup_steps=10)
     for batch in loader:
         loss = model(batch); loss.backward()
         opt.step()                    # eager for warmup_steps, then captured+replayed
         opt.zero_grad(set_to_none=False)   # MUST keep .grad tensors stable for replay
+        sched.step()
 
 Requirements / caveats:
+  * LR schedulers must be constructed on the *inner* optimizer: torch's ``LRScheduler``
+    rejects anything failing ``isinstance(opt, torch.optim.Optimizer)``, which this
+    wrapper is not. Scheduling still works -- the scheduler writes ``group["lr"]`` and the
+    wrapper pushes that into the device LR tensors before each replay.
   * ``loss.backward()`` must accumulate into the same ``p.grad`` tensors each step, so
     call ``zero_grad(set_to_none=False)`` (the wrapper enforces this). The first
     backward allocates ``.grad``; capture pins those buffers.
   * The step must not do a host sync (``.item()``/``.cpu()``) or change tensor shapes
     across steps. Dion2/NorDion2's selection *count* is fixed; only indices/values vary,
     which replay handles (it re-reads the live tensors).
+  * On the sharded path the graph holds the captured megabatch all-to-all, so the NCCL
+    ops outlive the step. ``dist.destroy_process_group()`` blocks while they are alive:
+    drop the wrapper (and any graph it holds) before tearing the process group down at
+    the end of a run, or shutdown hangs.
   * torch.compile is neither needed nor wanted under a graph (the graph already removes
     dispatch, and a fullgraph compile of the unrolled per-matrix loops is what makes the
     first step take minutes at many layers). Disable it for the wrapped optimizer.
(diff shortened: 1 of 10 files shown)


JohnLangford:
Closing in favor of #105.

The CUDA-graph approach here collapses host dispatch ~500×, but only with a **fixed** learning rate (the LR is baked at capture). Making it capturable under a real scheduled LR requires a device-tensor LR, and that regresses GPU (device) time **10–30%**: a device-scalar `scalar * update` multiply drops out of PyTorch's vectorized elementwise kernel into the general (non-vectorized) one — confirmed by profiling, and worse on the Triton write-back path. Since a fixed LR doesn't match actual training usage, the capturable path isn't worth that GPU cost.

#105 gets the host-dispatch reduction a different, cheaper way — batching the per-matrix scatter loops into stacked scatters — which is bit-exact, leaves GPU time and kernel vectorization untouched, and composes with a normal scheduled LR (no capturability constraint).

JohnLangford:
Reopening. The real-architecture optimizer-step timings for the paper show that on the megabatched distributed step, Dion2/Dion3 **filtering** is host-dispatch-bound at small per-GPU compute: its per-matrix selection + error-feedback bookkeeping launches ~1.8× the kernels of the unfiltered baseline (~8400 vs ~4600 at 14B/8-GPU), which in eager execution is a net *slowdown* below ~5B and leaves ~30–48 ms of host time on the table at 14B. Capturing the step and replaying it collapses that dispatch to a single launch — measured e.g. 1B Dion2-¼ 30.6 → 5.1 ms, and it makes filtering a per-GPU speedup at *every* scale (1.1× at 1B → 4.3× at 14B) instead of only at large scale.

Since filtering is the configuration we actually train with (best statistical performance), the capture path is worth having on by default. This branch is the drop-in wrapper; a companion PR against the training stack (`microsoft/aifsdk`, semantic/control) bumps the dion pin here and wraps the optimizer in `CudaGraphOptimizer` behind a config flag defaulting on.

Correctness for a **scheduled LR** is the one non-obvious requirement and is covered here: the matrix-path LR is carried as a persistent device tensor refreshed outside the graph (`_group_lr_tensor` / `_refresh_lr_tensors`), so a warmup-decay-cooldown schedule survives replay — verified bit-exact eager-vs-replay (`tests/test_cuda_graph.py`, plus a scheduled-LR convergence check: eager-vs-replay diverges 0.0 under an identical schedule).

PR #105, opened 2026-07-16
Batch the Dion2/NorDion2 megabatch per-matrix scatter loops (bit-exact host reduction)

JohnLangford:
## What

The megabatched Dion2/NorDion2 optimizer step ran **one scatter kernel per weight matrix per step** — once for the error-feedback momentum update (`dion2_pre_orthogonalize`) and once for the write-back (`dion2_post_orthogonalize`). That per-matrix host dispatch is a large share of the opt-step's host cost. This batches each per-matrix loop, within a shape group, into a **single stacked scatter**: stack `M`/`X`/`U`/`indices`, do one `scatter_`/`scatter_add_` over the batch, then a `foreach` copy-back.

The selected indices are unique per slice, so the batched scatter is collision-free and **bit-exact** with the per-matrix loop.

## Why this over a CUDA graph

A CUDA-graph capture collapses host dispatch far more (~500×), but only with a **fixed** learning rate — the LR gets baked at capture, and making it capturable (device-tensor LR) regresses GPU time 10–30% because a device-scalar `scalar * update` drops out of PyTorch's vectorized elementwise kernel. This batching needs none of that: it's a pure launch-count reduction, so it **composes with a normal scheduled LR** and leaves the GPU kernels (and their vectorization) untouched.

## Results

Single-GPU, dion2 f=½, 24 layers, opt-step measured in isolation:

| | GPU (device) | CPU (host dispatch) |
|---|---|---|
| per-matrix loop | 16.5 ms | 7.5 ms |
| **batched** | 16.2 ms | **4.3 ms** (~1.7×) |

GPU time unchanged (noise); bit-exact updates (fixed-seed fingerprint `-7937.4657929559` dion2, `-7934.6272033966` nordion2). The absolute host reduction grows with the number of matrices per shape group (larger at multi-B scale).

## Testing

Existing `tests/test_optimizers.py` (parity / determinism / megabatch-same-shape) and `tests/test_adamw_foreach.py` pass unchanged (102 passed) — the megabatch-vs-reference parity tests exercise the batched write-back directly.

### DIFF
--- a/dion/dion2.py
+++ b/dion/dion2.py
@@ -527,17 +527,15 @@ def dion2_pre_orthogonalize(
         )
         selected_stacked = torch.gather(M_stacked, dim=-1, index=indices_expanded)
 
-    # Apply error feedback decay to selected slices in the original M tensors.
-    # We reuse the already-gathered slices and write them back (scaled) using
-    # scatter_, which places values into positions specified by the index tensor.
+    # Apply error-feedback decay to the selected slices, batched over the whole shape
+    # group: one scatter into the stacked momentum plus a foreach copy-back, instead of
+    # one ``scatter_`` per matrix. The per-matrix loop was a large share of the opt-step
+    # host dispatch (~one scatter kernel per weight matrix per step); ``indices_expanded``
+    # already has the batched (N, k, cols)/(N, rows, k) shape, and the selected slices are
+    # unique per row/column, so this is bit-exact with the per-tensor loop.
+    M_stacked.scatter_(dim=select_dim, index=indices_expanded, src=selected_stacked * ef_decay)
+    torch._foreach_copy_(M, list(M_stacked.unbind(dim=0)))
     indices_list = list(indices.unbind(dim=0))
-    selected_list = list(selected_stacked.unbind(dim=0))
-    for m, idx, selected in zip(M, indices_list, selected_list):
-        if select_dim == -2:
-            idx_exp = idx.unsqueeze(-1).expand(*idx.shape, m.size(-1))
-        else:
-            idx_exp = idx.unsqueeze(-2).expand(*idx.shape[:-1], m.size(-2), idx.shape[-1])
-        m.scatter_(dim=select_dim, index=idx_exp, src=selected * ef_decay)
 
     # Convert to bf16 and unstack for communication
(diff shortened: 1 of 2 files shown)


JohnLangford:
Closing — same-run measurement shows this is a net loss at scale, not the free host win I claimed.

Batching the scatter requires `torch.stack(X)` + `foreach_copy_` back, i.e. two extra full copies of the shape-group's parameters per step, and that memory traffic scales with matrix width. Same-run single-GPU at 5120-wide matrices: stacked **254–259 ms GPU** vs per-matrix loop **245–246 ms** (~+4.7%), reproducible across three pairs. It's invisible at 1024-wide (why my earlier test looked neutral).

That collides with the regime: at 14B/8-GPU the opt-step is GPU-bound (host ~10 ms ≪ GPU ~80 ms), so this trades a *hidden* host saving (~8 ms, overlapped behind GPU work) for a *real* GPU cost (~4 ms, on the critical path) — net wall-clock worse where it matters. It'd only help in the host-bound corner. The in-place per-matrix loop is the correct default; keeping it.

JohnLangford:
Reopening, and restacking this on top of #104 (`jcl/cudagraph-capturable-optstep`).

Reversing the earlier decision to drop this. Benchmarking the capture path (#104) on the **real architecture** showed the per-matrix scatter loop is not just slower host-side under filtering — **it cannot be CUDA-graph-captured at all**. With #104 alone, every filtered Dion2/NorDion2 configuration (`f = 1/2`, `1/4`) aborts capture with `CUDA error: operation failed due to a previous error during capture`, at every scale (1B–14B), single-device and 8-GPU sharded alike; the unfiltered configurations capture fine. Filtering is the configuration we train with (best statistical performance), so capturing it is the whole point.

Batching the per-matrix `scatter_` / `scatter_add_` into one scatter over the stacked shape-group (this PR) makes the filtered step graph-safe — the hundreds of small per-matrix scatter kernels + temporaries that the loop feeds into a single capture are what fail. So #104 + #105 together are the shipping path; #104 is not sufficient on its own for filtered capture.

The ~4.7% GPU cost measured earlier is on the filtered *compute* and is dwarfed by capture's host savings (the filtered step drops ~6× at 1B once it can be captured), so with capture as the default the trade-off is clearly net-positive. #104's test suite never exercised the filtered path, which is why this stayed latent; a filtered-capture regression test belongs here.

PR #107, opened 2026-07-26
Batch the Dion2/NorDion2 megabatch per-matrix scatter loops (bit-exact host reduction)

JohnLangford:
Supersedes #105 (its head branch was rebased onto #104, so GitHub would not let it reopen). **Stacked on #104** (`jcl/cudagraph-capturable-optstep`): this branch contains #104's commits too, so the reviewable diff for *this* PR is the two scatter-batching commits on top — review #104 first.

## What

Batch the Dion2/NorDion2 megabatch per-matrix `scatter_` / `scatter_add_` loops into a single scatter over the stacked shape-group (bit-exact host reduction).

## Why this is required for CUDA-graph capture (not just perf)

Benchmarking #104's capture path on the real architecture showed the per-matrix scatter loop **cannot be CUDA-graph-captured**. With #104 alone, every filtered Dion2/NorDion2 configuration (`f = 1/2`, `1/4`) aborts capture with `CUDA error: operation failed due to a previous error during capture`, at every scale (1B–14B), single-device and 8-GPU sharded alike; unfiltered configurations capture fine. The hundreds of small per-matrix scatter kernels + temporaries the loop feeds into one capture are what fail; the batched scatter is graph-safe.

Filtering is the configuration we train with, so capturing it is the point — #104 + this PR together are the shipping path. The ~4.7% GPU cost is on the filtered compute and is dwarfed by capture's host savings (the filtered step drops ~6× at 1B once capturable). A filtered-capture regression test is included so this stays covered.

### DIFF
--- a/dion/cuda_graph.py
+++ b/dion/cuda_graph.py
@@ -12,19 +12,30 @@ graph launch (GPU time unchanged), and the capture holds through the megabatch a
 Usage (drop-in around any optimizer whose step() is graph-safe -- no host syncs, fixed
 shapes, gradients living in stable ``.grad`` tensors):
 
-    opt = CudaGraphOptimizer(Dion2(param_groups, ...), warmup_steps=10)
+    inner = Dion2(param_groups, ...)
+    sched = torch.optim.lr_scheduler.CosineAnnealingLR(inner, T_max=...)  # NOT the wrapper
+    opt = CudaGraphOptimizer(inner, warmup_steps=10)
     for batch in loader:
         loss = model(batch); loss.backward()
         opt.step()                    # eager for warmup_steps, then captured+replayed
         opt.zero_grad(set_to_none=False)   # MUST keep .grad tensors stable for replay
+        sched.step()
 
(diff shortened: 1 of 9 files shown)


JohnLangford:
Closing. My "required for filtered capture" rationale for this PR was wrong. Verifying #104's capture path on the combined branch showed the filtered configs still aborted capture *with* this fusion in place — so the per-matrix scatter was never the cause.

The real cause was a `.item()` host sync in `dion2_post_orthogonalize_triton` (passing the weight-decay factor and adjusted LR to the kernel as host scalars), which becomes a GPU→host sync once the LR is a device tensor and is forbidden mid-capture. Fixed in #104 by passing those scalars as 0-d device tensors the kernel loads. With that, filtered Dion2/NorDion2 capture bit-exact (regression test added there), so this batching isn't needed for capturability. It remains a valid standalone host-dispatch optimization, but it's not on the path to the default and carries a small GPU cost, so I'm not pursuing it now.

